PR #1, opened 2016-05-19
Initial structure with a Feature Pack and Distribution

ppalaga:


ppalaga:
@jpkrohling and @stefannegrea could you please review?

I activated Travis CI in settings, but it still does not start.

ppalaga:
Great, CI started.

ppalaga:
Great, CI is passing

.travis.maven.settings.xml
```diff
@@ -0,0 +1,100 @@
...
+  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
+                      http://maven.apache.org/xsd/settings-1.0.0.xsd">
+
+  <interactiveMode>false</interactiveMode>
+
+  <servers>
+    <server>
+      <id>jboss-snapshots-repository</id>
+      <username>${env.NEXUS_DEPLOY_USERNAME}</username>
+      <password>${env.NEXUS_DEPLOY_PASSWORD}</password>
```
jpkrohling: I don't think the build should contain any server listed. And if the password is set this way, there's no way to use the encryption features Maven provides, right?

.travis.maven.settings.xml
```diff
@@ -0,0 +1,100 @@
...
+    </server>
+  </servers>
+
+  <profiles>
+    <profile>
+      <id>securecentral</id>
+      <!--Override the repository (and pluginRepository) "central" from the
+         Maven Super POM -->
```
jpkrohling: Why? Isn't it an user choice? What if I'm using the Google mirror for the Central? 
https://maven-central.storage.googleapis.com

jpkrohling:
Looks good :) As long as it builds an usable distribution, I'm ok with this.

jmazzitelli:
I could not find where this was done, but I see this in the agent's storage-adapter config:

&lt;storage-adapter tenantId="${hawkular.rest.tenantId}" ...

Note that tenantID is not required, if not specified, we use the default "hawkular". I think you should at least allow for that default value if hawkualr.rest.tenantId is not specified, like this:

&lt;storage-adapter tenantId="${hawkular.rest.tenantId:hawkular}" ...

This would avoid having a person always specify a tenant on the command line.

How are people suppose to set these? On the "standalone.sh" command line via -D? If they forget, the agent will fail to startup and dump errors.

jmazzitelli:
> As long as it builds an usable distribution, I'm ok with this.

The agent is not configured with an OOTB username/password. It just has:

&lt;storage-adapter username="${hawkular.rest.user}" password="${hawkular.rest.password}" ...

Does this services build not have a default user anymore? If it does not, we'll need some really good docs to tell people that they can't just unzip and run and see Hawkular be in a usable state immediately. If there is a default user, we should default the agent to use those like this:

&lt;storage-adapter username="${hawkular.rest.user:jdoe}" password="${hawkular.rest.password:thePasswordHere}" ...

I would definitely prefer the latter - make it easy on people to download and go for demo and development purposes - even if it means having a -Pdev profile to build a pre-populated user in the distro (like we had in the hawkular kettle builds).

.travis.maven.settings.xml
ppalaga: > @jpkrohling I don't think the build should contain any server listed. And if the password is set this way, there's no way to use the encryption features Maven provides, right?

I did not think much about those ${env.NEXUS_DEPLOY_*} . I admit I blindly copied it from another component. Thinking of it now, I decided to remove the whole `<servers>` section, because (1) the snapshots are useless and (2) there is no snapshots deployment in `travis.yaml/after_success` anyway.

.travis.maven.settings.xml
ppalaga: > @jpkrohling Why? Isn't it an user choice? What if I'm using the Google mirror for the Central?
https://maven-central.storage.googleapis.com

The present `...settings.xml` file is supposed to be used by Travis and you are free to use any other settings in your local `settings.xml` that suit your needs. The main motivation for overriding the central here is to use `https` rather than the default `http`. I see no need to change anything here.

ppalaga:
> @jpkrohling Looks good :) As long as it builds an usable distribution, I'm ok with this.

I have not booted it once and as you certainly noticed there is no single itest (that we should add soon for sure). 

I also noticed that the Alert's JMX objects are missing in `hawkular-services-messaging-activemq.xsl` which needs to be fixed now.

ppalaga:
> I could not find where this was done, but I see this in the agent's storage-adapter config:
> 
> <storage-adapter tenantId="${hawkular.rest.tenantId}" ...
> 
> Note that tenantID is not required, if not specified, we use the default "hawkular". I think you should at least allow for that default value if hawkualr.rest.tenantId is not specified, like this:
> 
> <storage-adapter tenantId="${hawkular.rest.tenantId:hawkular}" ...

+1 I have just sent a fix to Agent: https://github.com/hawkular/hawkular-agent/pull/190

> This would avoid having a person always specify a tenant on the command line.
> 
> How are people suppose to set these? On the "standalone.sh" command line via -D?

Yes, -D was the intention.

> If they forget, the agent will fail to startup and dump errors.

Yes, I think "fail to startup and dump errors" should be expected for a non-dev build. 

A related question is if non-empty credentials should be required in the Agent's `storage-adapter` config, because Hawkular backend components can (in theory) be configured not to require any credentials.

ppalaga:
> The agent is not configured with an OOTB username/password. It just has:
> 
> <storage-adapter username="${hawkular.rest.user}" password="${hawkular.rest.password}" ...
> 
> Does this services build not have a default user anymore? If it does not, we'll need some really good docs to tell people that they can't just unzip and run and see Hawkular be in a usable state immediately. If there is a default user, we should default the agent to use those like this:
> 
> <storage-adapter username="${hawkular.rest.user:jdoe}" password="${hawkular.rest.password:thePasswordHere}" ...
> 
> I would definitely prefer the latter - make it easy on people to download and go for demo and development purposes - even if it means having a -Pdev profile to build a pre-populated user in the distro (like we had in the hawkular kettle builds).

Yes, there is no `dev` profile here ATM and I am going to add it now. Thanks for bringing this forward. Nevertheless, the zip that will be deployed to Maven Central and offered for download on hawkular.org will not have a hard coded users (same as hawkular zip).

ppalaga:
@jpkrohling @jmazzitelli 37102b6 : 

* Brings fixes I have promised above
* Has `dev` profile
* Adds caches and JMS topics and queues required by Alerts
* It boots without apparent failures
* Still no itests

Is this good to merge?

jmazzitelli:
I'm building this now and will test it with my latest agent. That should be a good test to see if this is ready to merge.

feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template.xsl
```diff
@@ -0,0 +1,60 @@
...
+  <xsl:template match="/*[local-name()='server']/*[local-name()='system-properties']/*[local-name()='property'][last()]">
+    <xsl:message>hawkular.rest.user = <xsl:value-of select="$hawkular.rest.user" /></xsl:message>
+    <xsl:copy>
+      <xsl:apply-templates select="node()|comment()|@*"/>
+    </xsl:copy>
+    <xsl:element name="property" namespace="{namespace-uri()}">
+      <xsl:attribute name="name">hawkular.agent.enabled</xsl:attribute>
+      <xsl:attribute name="value">${hawkular.agent.enabled:<xsl:value-of select="$hawkular.agent.enabled" />}</xsl:attribute>
+    </xsl:element>
+    <xsl:if test="$hawkular.rest.user != ''">
```
jpkrohling: If this property is not set, it's not added to standalone.xml? Wouldn't it be better to have it there all the time?

jpkrohling: Or is it just for the XSLT parameter itself?

ppalaga: Yes, `<xsl:if test="$hawkular.rest.user != ''">` is checking the `hawkular.rest.user` param passed to XSLT. If the param is set (which is the case only in `dev` Maven profile) then `<property name="hawkular.rest.user" value="myUser"/>` is added under `<system-properties>` of `standalone.xml`.

Also note, that no matter if `<property name="hawkular.rest.user" value="myUser"/>` element is there or not, the `${hawkular.rest.user}` property is referenced in the Agent subsystem config - i.e. if `<property name="hawkular.rest.user" value="myUser"/>` is not there, then `-Dhawkular.rest.user=myUser` must be set on command line.

In other words, `<property name="hawkular.rest.user" value="myUser"/>` is added in the `dev` Maven profile so that the developer does not need to set it on command line.

jpkrohling: > so that the developer does not need to set it on command line.

I'm more concerned about the end user not finding where to set this. But I actually prefer your current solution than any other solution I can think of for now. If this turns out to be a problem, we can change it later.

PR #2, opened 2016-05-25
HAWKULAR-1078 - First step in having SSL on by default

jpkrohling:


jpkrohling:
A few comments: 

- This is intended as a preview and pre-testing by component owners
- The final implementation would *probably* add the `keystore` only for -dev builds. This is not sorted out yet.
- HTTP2 is added to the listener, but more changes are required to test HTTP2.
- The cert in the keystore is self-signed, so, calls via `curl` have to include `-k` (or `--insecure`)

jpkrohling:
@ppalaga , @pilhuhn , could you please review? But please, don't merge yet! :)

ppalaga:
> The final implementation would probably add the keystore only for -dev builds. This is not sorted out yet.

Hardcoding the keystore makes the distro vulnerable, e.g. to a MITM attack, right?

feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template.xsl
```diff
@@ -50,6 +50,23 @@
...
+      <xsl:apply-templates select="node()|comment()|@*" />
+    </xsl:copy>
+    <server-identities>
+      <ssl>
+        <keystore
+            path="hawkular-server.keystore"
+            relative-to="jboss.server.config.dir"
+            keystore-password="changeme"
+            alias="hawkular-server"
+            key-password="changeme" />
```
ppalaga: What is the relationship of the `changeme` string to the hardcoded keystore? Was `changeme` used to encrypt it?

jmazzitelli:
In RHQ, we used to build the self-signed keystore on each build so at least it wasn't the same certificate each time. Not sure if we care but here's how we did it in RHQ:

https://github.com/rhq-project/rhq/blob/master/modules/enterprise/server/appserver/src/main/scripts/rhq-container.build.xml#L1114-L1123

feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template.xsl
jpkrohling: One is the password for opening the keystore. The other is the password for having access to the key itself (ie: our cert). So, a new keystore with new certs implies in new passwords, but those passwords can be changed without affecting the certificate itself.

jpkrohling:
> Hardcoding the keystore makes the distro vulnerable, e.g. to a MITM attack, right?

Yes, and it's actually even worse: shipping the keystore means that we are shipping the private key for the cert. So, anyone using these certs should consider that the cert is public. Data can be captured by a man in the middle.

> In RHQ, we used to build the self-signed keystore on each build so at least it wasn't the same certificate each time.

I'm not sure this brings any benefits. IMO, it's actually bad, as it gives a false sense of security. I'd rather ship with a cert that is valid only for localhost, meaning that it's broken for a wider usage.

jpkrohling:
This might now be obsolete, given that Wildfly 10.1 might have a similar feature built in: http://lists.jboss.org/pipermail/wildfly-dev/2016-June/005040.html

PR #4, opened 2016-05-26
Update Metrics hash to fix /status and /static/* requiring auth.

stefannegrea:


ppalaga:
Now after #3 was merged, there is a test checking `/metrics/status`: https://github.com/hawkular/hawkular-services/blob/master/itest/src/test/java/org/hawkular/services/rest/test/MetricsITest.java#L50

As it statnds now, the client send an Auth header. Should we not change the test to send the request without Auth header?

jpkrohling:
Merge after #8

jpkrohling:
Actually, I'll close this one. Moving to Metrics `0.16.0` would be a better option at this point.

ppalaga:
@jpkrohling , FYI, me and @stefannegrea agreed that the itests in services accessing `/status` should do so without both Auth and Tenant headers.

jpkrohling:
Absolutely, it makes sense.

PR #5, opened 2016-05-26
Smoke itests for Alerting

lucasponce:


ppalaga:
@lucasponce excellent, thanks!

PR #8, opened 2016-06-01
Imported REST from Hawkular to Services

jpkrohling:


jpkrohling:
This is not compiling, due to the security integration still depending on Accounts. @metlos , @Jiri-Kremser , could you take a look and align it with the setup that Inventory has?

jpkrohling:
By the way: only the last two commits are interesting for a review. The third-to-last is a merge commit, and everything before that was imported from the previous repository.

metlos:
Hmm.. what version of inventory does this depend on? I thought `0.16.0.Final` has no direct dependency on accounts.

metlos:
Ok, looking at the maven output I think I delegate this to @Jiri-Kremser :smile: This seems to be coming from the `hawkular-rest-api` module that seems to be consuming Inventory's REST API as a compile-time dependency and assumes classes that are no longer there...

jpkrohling:
It's using `0.16.0.Final-SRC-revision-95d9738893fd0a536ca08919ff0a512dad2a90db` , but note that this is not about Inventory itself, but `rest`. In some places, it uses a similar logic as the one that inventory used to have. So, it needs to be changed to the logic that inventory currently has.

jpkrohling:
From what I could see, the whole security integration purpose is to have a Tenant ID (besides the startup tasks, like creating resources). If that's the case, then the solution looks quite simple: just remove everything and `@Inject @TenantId private String tenantId` on `RestBase`, no?

jpkrohling:
Doing what I mentioned on the comment above got the build working, but the following can be seen on the server boot:

```
11:43:30,520 ERROR [org.jboss.as.controller.management-operation] (Controller Boot Thread) WFLYCTL0013: Operation ("add") failed - address: ([("deployment" => "hawkular-rest-api.war")]) - failure description: {"WFLYCTL0080: Failed services" => {"jboss.module.service.\"deployment.hawkular-rest-api.war\".main" => "org.jboss.msc.service.StartException in service jboss.module.service.\"deployment.hawkular-rest-api.war\".main: WFLYSRV0179: Failed to load module: deployment.hawkular-rest-api.war:main
    Caused by: org.jboss.modules.ModuleNotFoundException: org.hawkular.commons.guava:main"}}
```

jpkrohling:
```
$ curl -u jdoe:password http://localhost:8080/hawkular/api/ping ; echo
Thu Jun 02 12:08:31 CEST 2016
```

It seems the most basic test works. Again, the last three commits are the relevant ones. For a review, it might be better to see the diff on the CLI, as github is a bit confused about the merge commits. After pulling the changes from this PR into a local branch, run a diff against master:

`git diff master`

jpkrohling:
Ready for re-review. @metlos , @Jiri-Kremser , whenever you want :) Please, pay special attention to the security changes on the last three commits: basically, I did what I mentioned on https://github.com/hawkular/hawkular-services/pull/8#issuecomment-223244313

hawkular-rest/hawkular-rest-api/pom.xml
```diff
@@ -0,0 +1,201 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>hawkular-inventory-json-helper</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>hawkular-inventory-security-permissive</artifactId>
```
metlos: The way this works in inventory is that the REST API itself depends merely on the `hawkular-inventory-security-spi` and the "provisioning" of the implementation of it is left for the distribution composition, i.e. `hawkular-inventory-dist`. While I'm not against using a simpler approach here, or even in inventory itself, by just using permissive security everywhere, it should be at least documented that all the places that declare the SPI impl should use the same impl (IMHO permissive is not what we will be using in the end)...

hawkular-rest/hawkular-rest-api/pom.xml
```diff
@@ -0,0 +1,201 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.logging</groupId>
+      <artifactId>jboss-logging-processor</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <!-- docs -->
+    <dependency>
+      <groupId>com.wordnik</groupId>
```
metlos: isn't `io.swagger` what we should be using now (as it's newer)? @ppalaga, could you please comment on which one of the swagger versions we should be using?

hawkular-rest/hawkular-rest-api/src/main/java/org/hawkular/rest/PageToStreamThreadPool.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.annotation.PostConstruct;
+import javax.enterprise.inject.Produces;
+import javax.inject.Singleton;
+
+/**
+ * @author Jirka Kremser
+ * @since 0.3.4
+ */
+@Singleton
+public class PageToStreamThreadPool {
```
metlos: This is copied verbatim from inventory (including the `@since` tag :) ). Shouldn't this be moved somewhere where we can share a single impl of this class?

hawkular-rest/hawkular-rest-api/src/main/java/org/hawkular/rest/RequestUtil.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.hawkular.inventory.api.paging.Order;
+import org.hawkular.inventory.api.paging.PageContext;
+import org.hawkular.inventory.api.paging.Pager;
+
+/**
+ * @author Lukas Krejci
+ * @since 0.0.1
+ */
+public class RequestUtil {
```
metlos: Again, verbatim copy of stuff in inventory.

hawkular-rest/hawkular-rest-api/src/main/java/org/hawkular/rest/ResponseUtil.java
```diff
@@ -0,0 +1,191 @@
...
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SequenceWriter;
+
+/**
+ * @author Lukas Krejci
+ * @author Heiko W. Rupp
+ * @since 0.0.1
+ */
+final class ResponseUtil {
```
metlos: This is almost the same as a class of the same name in inventory. One more reason to put such classes into a common shared artifact to avoid duplicating bugs/fixes..

hawkular-rest/hawkular-rest-api/src/main/java/org/hawkular/rest/RestPing.java
```diff
@@ -0,0 +1,175 @@
...
+import org.jboss.resteasy.core.ResourceMethodRegistry;
+
+/**
+ * @author Lukas Krejci
+ * @since 1.0
+ */
+@Path("/")
+@Produces(value = APPLICATION_JSON)
+@Consumes(value = APPLICATION_JSON)
+public class RestPing {
```
metlos: Again, copied from inventory.

hawkular-rest/hawkular-rest-api/src/main/java/org/hawkular/rest/ResponseUtil.java
jpkrohling: Agree, but this is how it was on the original hawkular-rest.war module. Note that I moved the module from the main Hawkular repository to this repository, and changed only the auth part.

hawkular-rest/hawkular-rest-api/pom.xml
jpkrohling: How would that work in the context of this module? It builds a WAR itself, so, it has to include an implementation, right?

metlos: Not sure - we can maybe just document it here, that the impls need to be in sync, or we could maybe do it on Wfly level somehow using modules, if possible. Not sure if that would be doable using CDI or maybe we would have to switch to using META-INF services and "persuade" jboss-modules to discover the impl of the SPI amongst the modules. Not sure if that is worth the effort though. So maybe a comment here would be enough?

jpkrohling: What are the possible modules, and in which cases are they going to be used? If there would be a standalone inventory + inventory on Hawkular Services, then there could be a meta-module `hawkular-inventory-security-services`, which then depends on the appropriate module for the services. Should the implementation change, just change the transitive dependency, instead of requiring all consumers to use the new one.

metlos: Right now, we have permissive and accounts impls of the SPI. Inventory REST API uses the SPI on every request to check that the user has CRUD privileges on entities being accessed. Not sure what the Hawkular REST API uses it for exactly.. If it is a read-only interface, then really it would be using the SPI only to get the tenant ID.

metlos:
The changes to the security aspect of this PR look ok to me. I'll leave the decision about what to do with permissive to you. I'm ok with both documenting it or hiding the impl behind a "meta-module" with the latter being a better yet more laborious approach, obviously.

jpkrohling:
As this is a complex PR, moving the history from one place to another, I'd merge this and add JIRA for the following tasks:
- add a meta-module on Inventory, that would add dependency on `permissive`
- remove the comment from `RestBase`
- evaluate the `swagger` dependency
- refactor the duplicate code

hawkular-rest/hawkular-rest-api/pom.xml
ppalaga: > isn't io.swagger what we should be using now (as it's newer)? @ppalaga, could you please comment on which one of the swagger versions we should be using?

Yes, `io.swagger` is newer and better than `com.wordnik`.

PR #10, opened 2016-06-02
Test metrics collection against /metrics/gauges/{id}/raw rather than /metrics/ga

ppalaga:


ppalaga:
@jpkrohling could you please review and eventually merge?

jpkrohling:
Looks good, but I'll merge after #8 , otherwise I might have problems with the merge commit on that PR.

PR #14, opened 2016-06-07
Add status check on the front page

theute:
The status for alerts will fail until https://issues.jboss.org/browse/HWKALERTS-148 is fixed

untitled


jpkrohling:
Looks nice!

PR #18, opened 2016-06-16
Added tar.gz to packaging

jpkrohling:


jpkrohling:
@abonas , @pilhuhn : we had .tar.gz for the main Hawkular distribution, but somehow it was not added to services. If that was not a deliberate decision, this PR re-introduces this format (which is nicer on Linux environments)

jpkrohling:
@ppalaga : do you miss Hawkular? :) If so, you are also free to review this!

jpkrohling:
For server optimized Fedora/RHEL/CentOS (ie: minimal installations), `unzip` is not available. So, for automated deployments, it's useful (and easier) to have `.tar.gz` as well. Not to mention the `--strip` parameter for `tar`, which is not available for `zip`, meaning that we have to do some bash magic to move from `hawkular-services-cd2/hawkular-services-dist-0.0.3-SNAPSHOT` to `hawkular-services-cd2`.

jpkrohling:
If we had to choose between `tar.gz` and `zip`, I'd keep `tar.gz`, as I'm not even sure -services work on Windows :)

ppalaga:
You persuaded me :)

ppalaga:
Feel free to merge in which ever way you like

PR #20, opened 2016-06-17
Hawkular-1083 Server Side Feed Availability Backfil

jshaughn:
- Monitor feed "pings" and take action on feeds that stop reporting:
  - Set to DOWN the feed's ping availability metric
  - Set to UNKNOWN the availability metrics being sent for
    the resources monitored by the feed.
- This is always for a specific tenant because a feed can be
  registered for multiple tenants.
- This is a general feature for feeds but is initially targeted
  for our agent, which is being updated to send avail pings
  to leverage this feature.


hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
```diff
@@ -0,0 +1,539 @@
...
+        if (standalone) {
+            log.info("Initializing Standalone Availability Cache");
+        } else {
+            log.info("Initializing Distributed Availability Cache");
+            processTopologyChange();
+        }
+
+        // This is basically a fixed size pool, the size may need to be increased if there are a lot of
+        // active feeds.
+        executorService = Executors.newScheduledThreadPool(JOB_THREADS);
```
jmazzitelli: I have to ask, only because it tripped my OCD. Why do you use /* ... */ comments only ten lines above, but yet use // here?

jshaughn: Good question, that was likely something I stole and didn't update.  I'll change that to be consistently //

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
```diff
@@ -0,0 +1,539 @@
...
+        Address member = cacheManager.getAddress();
+
+        if (null == members || null == member || -1 == members.indexOf(member)) {
+            log.error("Unexpected Cache Topology. Member: " + member + " not found in " + members);
+            return;
+        }
+
+        Collections.sort(members);
+        numMembers = members.size();
+        memberNumber = members.indexOf(member);
```
jmazzitelli: you are going to have to make access to these  memberNumber / numMembers thread safe. Because this can get triggered anytime in the listener - which may be at the same time "isResponsible" is called. So it seems to me it could be possible concurrency is going to potentially make you think this node isn't responsible for some metric when it is.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
```diff
@@ -0,0 +1,539 @@
...
+    private int numMembers = 1;
+
+    /**
+     * Each member assigned a unique number from 0..numMembers-1.
+     */
+    private int memberNumber = 0;
+
+    private ScheduledExecutorService executorService;
+
+    private Map<CacheKey, ScheduledFuture<?>> jobMap = new HashMap<>();
```
jmazzitelli: not sure, but I think this should be of ConcurrentHashMap in case you try to get and put at the same time (I see two methods accessing this)

jmazzitelli: Yeah, I think there is a chance for concurrent access - you might cancel a job while backfilling at the same time a metric comes in and you need to put stuff in.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
jmazzitelli: I see the self. references but where are the Lock annotations? I don't see anything that actually protects from concurrency.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
```diff
@@ -0,0 +1,540 @@
...
+
+        Collections.sort(members);
+        numMembers = members.size();
+        memberNumber = members.indexOf(member);
+
+        log.info("Topology Update. Member " + member + " assigned number " + memberNumber + " of " + numMembers);
+    }
+
+    @Override
+    public boolean isResponsible(String metricId) {
```
jmazzitelli: this uses numMembers, so should be using read-Lock annotations

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
```diff
@@ -0,0 +1,540 @@
...
+        int hash = metricId.hashCode();
+        int mod = hash % numMembers;
+        boolean result = mod == memberNumber;
+        log.trace("Member " + memberNumber + (result ? " is " : " is not ") + " responsible for " + metricId);
+        return result;
+    }
+
+    @Override
+    public void updateFeedAvailability(String tenantId, String feedAvailabilityMetricId) {
+        if (!isResponsible(feedAvailabilityMetricId)) {
```
jmazzitelli: should reference with "self" to ensure locking is done

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
```diff
@@ -0,0 +1,540 @@
...
+                        jobMap.put(key, sf);
+                    } else {
+                        log.debugf("Ignoring Backfill Job for %s, ping period %d > %d (the minimum)",
+                                key, pingPeriodMs, PING_PERIOD_MIN_SECS);
+                    }
+                }
+
+                // Update the cache with the latest ping
+                value.setLastUpdateTime(now);
+                backfillCache.put(key, value);
```
jmazzitelli: is this backfillCache thread-safe?

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
jshaughn: As I understand it, Singleton EJBs have implied Lock(WRITE) on all local methods.  So, I'm not sure we need them.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
jshaughn: It's already Write-locked, by default, should we change it?

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
jshaughn: I don't think we want to do that because we're already inside a Write-locked method.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
jshaughn: Uhm, I'm not sure about that, that's the ISpan cache, let me look into that.

jshaughn: OK, yes, Ispan Cache is a ConcurrentMap, so I think we can reduce updateFeedAvailability() to a READ lock.  Freeing up isResponsible() and updateFeedAvailability() should improve throughput as they will likely be the most called methods.

PR #22, opened 2016-06-20
put embedded cassandra in hawkular services if built with embeddedc profile

jmazzitelli:
This reintroduces the ability to build Hawkular Services with Embedded Cassandra.

You can do this via -Pembeddedc Maven profile.

This feature allows an easier "on ramp" for you developers or folks that want to demo - no need to install and start a separate Cassandra server (no need to even know that you have to do it).


pilhuhn:
I know that there have been discussions about that in both directions. 
I am merging this now purely for development purposes. I run myself with an external C* instance - even on a different IP.

PR #25, opened 2016-06-22
Dockerize hawkular-services

pilhuhn:
This PR introduces a way to create and start a Docker container with Hawkular services inside.

The current state to get it to work is

    $ mvn clean install
    $ cd docker-dist
    $ mvn docker:build docker:run

The created container will have a user of jdoe/password and the agent enabled.


jmazzitelli:
Can you get this to optionally disable the agent?

dist/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+MAINTAINER Heiko W. Rupp <hrupp@redhat.com>
+
+ADD target/hawkular-services-dist-0.0.2.Final-SNAPSHOT.zip /opt/hawkular.zip
+
+USER root
+
+env HAWKULAR_USER=jdoe
+env HAWKULAR_PASSWORD=password
+
```
jmazzitelli: env HAWKULAR_AGENT_ENABLED=true

dist/startcmd.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+/opt/hawkular/bin/standalone.sh -b 0.0.0.0 \
+       -bmanagement 0.0.0.0 \
+       -Djboss.server.data.dir=/opt/data \
+       -Dhawkular.rest.user=${HAWKULAR_USER} \
+       -Dhawkular.rest.password=${HAWKULAR_PASSWORD}
```
jmazzitelli: -Dhawkular.agent.enabled=${HAWKULAR_AGENT_ENABLED}

docker-dist/pom.xml
```diff
@@ -0,0 +1,295 @@
...
+
+  <artifactId>hawkular-services-dist-docker</artifactId>
+  <packaging>pom</packaging>
+
+  <properties>
+    <docker-plugin.project.version>0.15.6</docker-plugin.project.version>
+    <docker.base-image>jboss/wildfly:10.0.0.Final</docker.base-image>
+    <plexus-utils.version>3.0.15</plexus-utils.version>
+    <alpn-boot.version>8.1.7.v20160121</alpn-boot.version>
+    <version.cassandra>3.7</version.cassandra>
```
jmazzitelli: <hawkular.agent.enabled>true</hawkular.agent.enabled>

docker-dist/pom.xml
```diff
@@ -0,0 +1,295 @@
...
+                </ports>
+                <volumes>
+                  <volume>/opt/data</volume>
+                </volumes>
+                <user>root</user>
+                <env>
+                  <HAWKULAR_BACKEND>cassandra</HAWKULAR_BACKEND>
+                  <CASSANDRA_NODES>myCassandra</CASSANDRA_NODES>
+                  <HAWKULAR_USER>jdoe</HAWKULAR_USER>
+                  <HAWKULAR_PASSWORD>password</HAWKULAR_PASSWORD>
```
jmazzitelli: &lt;HAWKULAR_AGENT_ENABLED>${hawkular.agent.enabled}&lt;/HAWKULAR_AGENT_ENABLED>

docker-dist/src/main/resources/startcmd.sh
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+/opt/jboss/wildfly/bin/standalone.sh -b 0.0.0.0 \
+       -bmanagement 0.0.0.0 \
+       -Djboss.server.data.dir=/opt/data/data \
+       -Djboss.server.log.dir=/opt/data/log \
+       -Dhawkular.agent.enabled=true \
```
jmazzitelli: -Dhawkular.agent.enabled=${HAWKULAR_AGENT_ENABLED} \

jmazzitelli:
When you start the docker container, this gets run:

  /opt/hawkular/bin/standalone.sh -b 0.0.0.0 -bmanagement 0.0.0.0  ...

Does this prohibit the ability to run additional WildFly servers on your machine on different IPs like "127.0.0.5"? Rather than launch with different ports, I just always launch my servers on different IPs, that way all the ports always remain the same (8080, 9999, 9990, etc) - so I just have to remember 1 new number (the last digit of the new local IP - "5" in my example) rather than several new port numbers that are all offset from the other servers).

If not, can we make this configurable to be able to set the binding address with the default still 0.0.0.0? Make it configurable like the others (e.g. like the new HAWKULAR_AGENT_ENABLED I just commented about above).

docker-dist/pom.xml
```diff
@@ -35,6 +35,7 @@
     <plexus-utils.version>3.0.15</plexus-utils.version>
     <alpn-boot.version>8.1.7.v20160121</alpn-boot.version>
     <version.cassandra>3.7</version.cassandra>
+    <hawkular.agent.enabled></hawkular.agent.enabled>
```
jmazzitelli: you should set this to "true" rather than an empty string.

pilhuhn:
On 22 Jun 2016, at 14:34, John Mazzitelli wrote:

> When you start the docker container, this gets run:
>
>   /opt/hawkular/bin/standalone.sh -b 0.0.0.0 -bmanagement 0.0.0.0  ...
>
> Does this prohibit the ability to run additional WildFly servers on 
> your machine on different IPs like "127.0.0.5"? Rather than launch 
> with

No. Those bindings are strictly internal to the container.
What is visible outside is at the end determined by the -p parameters
on docker run.
Or the ports section in docker-compose.yml
   ports:
     - "8080:8080"
     - "8443:8443"
     - "9990:9990"

Or this in the pom

                 <ports>
                   <port>8080</port>
                   <port>8443</port>
                 </ports>

I am not sure how to set the bind addr of the container, but from the
docs it looks like --ip=<ip> would do it.

-- 
hwr@pilhuhn.de  m:0179/207 4919  b:http://pilhuhn.blogspot.com

jmazzitelli:
I followed the instructions:
```
$ mvn clean install
$ cd docker-dist
$ mvn docker:build docker:run
```

But during the mvn docker:build, I get this error:

```
[ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.15.6:build (default-cli) on project hawkular-services-dist-docker: Execution default-cli of goal io.fabric8:docker-maven-plugin:0.15.6:build failed: No url given, no DOCKER_HOST environment variable and no read/writable '/var/run/docker.sock' -> [Help 1]
```

rhuss:
You need to expose a `DOCKER_HOST` variable for accessing the Docker daemon. When using docker-machine use an `eval $(docker-machine env)` before.

jmazzitelli:
This brings up two points then:

1) you must have a docker daemon running (which I believe requires root)
2) I guess for the maven build, we have to assume the host to the daemon is "localhost" if we are to set DOCKER_HOST to some reasonable default.

jmazzitelli:
On Fedora 23. I export DOCKER_HOST in my shell (setting it to tcp://127.0.0.1). I have a docker daemon running as root. When I run mvn docker:build, I get this:

[ERROR] Failed to execute goal io.fabric8:docker-maven-plugin:0.15.6:build (default-cli) on project hawkular-services-dist-docker: Unable to pull 'jboss/wildfly:10.0.0.Final': <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
[ERROR] <html><head>
[ERROR] <title>404 Not Found</title>
[ERROR] </head><body>
[ERROR] <h1>Not Found</h1>
[ERROR] <p>The requested URL /v1.18/images/create was not found on this server.</p>
[ERROR] </body></html> (Not Found: 404)

docker-dist/pom.xml
```diff
@@ -0,0 +1,300 @@
...
+                <env>
+                  <CASSANDRA_START_RPC>true</CASSANDRA_START_RPC>
+                </env>
+                <volumes>
+                  <bind>
+                    <volume>/tmp/opt/hawkular/cassandra:/var/lib/cassandra</volume>
+                  </bind>
+                </volumes>
+              </run>
+            </image>-->
```
Jiri-Kremser: ><(((°>

docker-dist/pom.xml
```diff
@@ -0,0 +1,300 @@
...
+                <cmd>
+                  <shell>/opt/hawkular/bin/startcmd.sh</shell>
+                </cmd>
+                <assembly>
+                  <basedir>/</basedir>
+                  <descriptor>docker-assembly.xml</descriptor>
+                </assembly>
+                <runCmds>
+                  <run>ls -lsa ${docker.as.root}/standalone/configuration</run>
+                  <rum>whoami</rum>
```
Jiri-Kremser: :beer:

docker-dist/pom.xml
```diff
@@ -0,0 +1,300 @@
...
+                  <run>${docker.as.root}/bin/add-user.sh -a -u ${HAWKULAR_USER} -p ${HAWKULAR_PASSWORD} -g read-write,read-only</run>
```
Jiri-Kremser: this will end up in having 3 different RUN directives in a resulting docker file, which creates 3 different layers when building it => it'll increase the resulting size

a good practice is doing `RUN foo && bar && baz`

PR #32, opened 2016-07-01
Update inventory to 0.17.1.Final

metlos:


metlos:
Note that the itests still fail for me, due to failures with agent tests. That's because agent has not been updated in services to either use the new API or use the deprecated endpoint on its new location.

metlos:
Closing this in favor of #34.

PR #34, opened 2016-07-04
Inventory 0.17.2.final + Agent 0.20.0.final

metlos:
Doing these two commits as a single PR so that we see the two coming together in a green build.


pilhuhn:
Thanks @metlos

PR #38, opened 2016-07-19
Build docker images for hawkular-services

pilhuhn:
Allow to build docker images from Hawkular-services.
They need an external cassandra.


pgier:
This worked for me, (using Fedora 23).  From the docker-dist directory, run "mvn docker:build" and this will create the image and add it to the local registry.  Then I was able to "docker run" the new hawkular-services image based on wildfly.  To get it fully functioning, you also need to start up a cassandra instance and pass the ip/hostname via an environment var.

docker-dist/pom.xml
```diff
@@ -0,0 +1,214 @@
...
+          <watchInterval>500</watchInterval>
+          <logDate>default</logDate>
+          <verbose>true</verbose>
+          <autoPull>on</autoPull>
+          <watchPostGoal>org.apache.maven.plugins:maven-help-plugin:help</watchPostGoal>
+          <images>
+            <!-- Docker Image to use -->
+            <image>
+              <alias>hawkular-services</alias>
+              <name>${maven.username}/hawkular-services</name>
```
pgier: By default, the property ${maven.username} was not set for me, so I had to pass "-Dmaven.username=foo".  Should this use ${env.USERNAME} instead?

PR #39, opened 2016-07-28
Hawkular-1101 Backfiller does not honor DOWN availability

jshaughn:
Start listening for new FeedWebSocketClosedEvent messages on
HawkularCommandTopic and when detected force an avail backfill
for the down feed.

Also:
- convert some hardcoded literals to constants
- run the backfill job more often, default from 30s to 15s


jshaughn:
@jmazzitelli Please review but don't merge, it depends on commons snapshot, so that PR will need to be merged first.  Also, waiting for @pilhuhn to add any comments on the Jira.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/CommandEventListener.java
```diff
@@ -56,10 +59,14 @@
...
     private AlertsService alerts;
 
+    @EJB
+    BackfillCache backfillCacheManager;
+
     @Override
     protected void onBasicMessage(BasicMessage msg) {
 
         if (msg instanceof DeployApplicationResponse) {
+            // TODO: This may not work in a clustered env, I think it would execute N times, once per cluster member
```
jmazzitelli: if you put the message on a queue rather than topic, won't this problem go away?

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/CommandEventListener.java
```diff
@@ -83,10 +90,17 @@ protected void onBasicMessage(BasicMessage msg) {
             } catch (Exception e) {
                 log.errorf("Error processing event message [%s]: %s", msg.toJSON(), e);
             }
+        } else if (msg instanceof FeedWebSocketClosedEvent) {
```
jmazzitelli: to avoid a growing list of if-then-else statements, shouldn't we just have multiple MDB classes?

jmazzitelli:
LGTM

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/CommandEventListener.java
jshaughn: This comment was just an unrelated thing I noticed and not related to the PR.  But yes, it would go away because the first server to take the message would do the work and that would be that.  But, making this a queue would also defeat the purpose of the bus, I think, which is to broadcast happenings.  For example, the websocket feed closure message does need to be on a topic, because the same server should be processing the same feed repeatedly.  We could, I suppose, have a queue and a topic and let the sender decide where a particular message should go.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/CommandEventListener.java
jshaughn: The list has grown to 2, that doesn't seem too bad :)  But are you suggesting that we use a message selector approach with different MDBs?  It's doable.

PR #43, opened 2016-07-30
Bump versions of agent, commons and metrics.

pilhuhn:
cc @jshaughn @jpkrohling @jmazzitelli @stefannegrea


jpkrohling:
Looks good to me, but it looks like Metrics is still at Commons  0.7.1.Final.

pilhuhn:
@jpkrohling Is that an issue? I don't think metrics will bump that and re-release

jpkrohling:
Ideally, we would ship all the same versions. But I guess it's fine, at least for tomorrow's release.

PR #44, opened 2016-08-01
Upgrade Alerts 1.1.2.Final-SRC version

lucasponce:
@jshaughn Updated the alerts component with srcdeps to test the parameters.
I will update the JIRA with minimal documentation for the config parameters.


pilhuhn:
SRC-deps are a bit of a pain, as the maven release plugin (with the release 
profile) can't grok them.

Juca can tell more


Am 1. August 2016 5:38:39 nachm. schrieb Lucas Ponce 
<notifications@github.com>:

jshaughn:
> SRC-deps are a bit of a pain

We've been using them historically to avoid snapshot non-determinism.  Is there some other approach we should be using?

pilhuhn:
On 1 Aug 2016, at 19:39, Jay Shaughnessy wrote:

>> SRC-deps are a bit of a pain
>
> We've been using them historically to avoid snapshot non-determinism.  
> Is there some other approach we should be using?

Yes, full-releases of dependencies (alerts in this case). Just on demand 
when
you have changes that should go into -services.

jshaughn:
OK, after the fix is validated by Filip I'll cut an Alerting release and update Services to a release version.  We did this so that Filip could evaluate a probable fix w/o having to cut multiple releases.

PR #46, opened 2016-08-02
Hawkular-1104 Multiple Event Issue

jshaughn:
- Continue publishing command events to HawkularCommandEvent but convert
  it to a queue as opposed to a topic, limiting message consumption to a
  single handler.
- Change publishing FeedWebSocketClosed events from HawkularCommandEvent
  to a new topic, HawkularTopic.  This improves separation of concern,
  and maintains topic publishing for this event message.
- Add a new listener for HawkularTopic that is the new home for
  handling of FeedWSClosedEvent messages


jshaughn:
**Do Not Merge**
Snapshot Dependency in Place

PR #47, opened 2016-08-03
Add a global /status for Hawkular Services

lucasponce:
image
I have added a global /status endpoint for Hawkular Services version.
It will be helpful to check which version is running, avoiding to check the bytes.
Logic for the main STATUS state is an and of metrics,inventory and alerts status.
There are things that can be done in following PR like unify with inventory the same format of status, and extracting the manipulation of Manifest in common.
But I think this it could be a start point for adding this feature.
Please @jpkrohling can you start reviewing it ?
Thanks


hawkular-status/src/main/java/org/hawkular/services/rest/HawkularServicesApp.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author Jay Shaughnessy
+ * @author Lucas Ponce
+ */
+@ApplicationPath("/")
+public class HawkularServicesApp extends Application {
+    private static final Logger log = Logger.getLogger(HawkularServicesApp.class);
+
+    public HawkularServicesApp() {
+        log.debug("Hawkular Services started...");
```
jpkrohling: This is ambiguous. It might give the impression that the whole server has started, which is not true. It could be changed to "Hawkular Services Status Reporter started."

hawkular-status/src/main/java/org/hawkular/services/rest/ManifestUtil.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Manifest extraction.
+ * Credits to Hawkular Metrics team.
+ *
+ * @author Jay Shaughnessy
+ * @author Lucas Ponce
+ */
+@ApplicationScoped
+public class ManifestUtil {
```
jpkrohling: There's a class like that on commons. Why not use that one? 
https://github.com/jpkrohling/hawkular-commons/blob/master/hawkular-rest-status/src/main/java/org/hawkular/commons/rest/status/ManifestUtil.java

hawkular-status/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,26 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
+         version="3.1">
+
+  <display-name>Hawkular Services</display-name>
```
jpkrohling: Hawkular Services Status Reporter

hawkular-status/src/main/java/org/hawkular/services/rest/StatusHandler.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private static final Logger log = Logger.getLogger(StatusHandler.class);
+
+    @Inject
+    ManifestUtil manifestUtil;
+
+    @GET
+    @Path("/")
+    @Produces(APPLICATION_JSON)
+    public Response status(@Context ServletContext servletContext, @Context HttpServletRequest request) {
```
jpkrohling: To me, the best option would be to have the UI to query each service on its own. Adding new services would mean changing the UI only, instead of changing this endpoint.

hawkular-status/src/main/java/org/hawkular/services/rest/StatusHandler.java
```diff
@@ -0,0 +1,96 @@
...
+        Map<String, String> status = new HashMap<>();
+        status.putAll(manifestUtil.getFrom(servletContext));
+
+        String baseUrl = request.getRequestURL().toString();
+        String metricsUrl = baseUrl.replace("status", "metrics/status");
+        String inventoryUrl = baseUrl.replace("status", "inventory/status");
+        String alertsUrl = baseUrl.replace("status", "metrics/status");
+
+        Client baseClient = ClientBuilder.newClient();
+        Response metricsResponse = baseClient.target(metricsUrl).request("application/json").get();
```
jpkrohling: If querying each service is the approach to go, then I'd make parallel calls, instead of serial.

hawkular-status/src/main/java/org/hawkular/services/rest/ManifestUtil.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String IMPLEMENTATION_VERSION = "Implementation-Version";
+    private static final String BUILT_FROM_GIT = "Built-From-Git-SHA1";
+
+    private static final List<String> VERSION_ATTRIBUTES = ImmutableList.of(IMPLEMENTATION_VERSION,
```
jpkrohling: The dependency on guava is not listed on the pom. Besides, do we need guava (a few MBs) only for this? Isn't there another way to accomplish the same, without guava?

jpkrohling:
First round of review finished.

hawkular-status/src/main/java/org/hawkular/services/rest/StatusHandler.java
lucasponce: Not querying components status from this endpoint could be an option and simplify the code.
Point is then if the meaning of the status for the hawkular-services.
For me important data to extract is the versioning of the hawkular-services, querying and status is an extra that came thinking if we want to avoid querying per component status.
I can remove it if it doesnt make sense.

hawkular-status/src/main/java/org/hawkular/services/rest/ManifestUtil.java
lucasponce: Oh, I didn't notice it was there.
Indeed is even the same class we used.

hawkular-status/src/main/java/org/hawkular/services/rest/ManifestUtil.java
lucasponce: Yes, I think here a simple list is worth to not bring the dependencies.
But that means to modify commons first.

lucasponce: Yes, I've noticed that the class in commons removed the legacy dependency with guava.
What it makes sense.

hawkular-status/src/main/java/org/hawkular/services/rest/StatusHandler.java
jpkrohling: The main purpose of this /status is to inform the version of h-services. As such, this module can be almost empty, if it consumes the status handler from commons. I don't see anything wrong with that :)

lucasponce: Alright, I will simplify it. Yes main purpose is the version of h-services.

lucasponce:
@jpkrohling removed the composition of status logic and reused the status library from commons.

lucasponce:
I miss one thing, I guess it is useful to show the commit on the splash page, too.
That is missing I'm going to add it if.

hawkular-status/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+    </dependency>
+
+    <!-- docs -->
+    <dependency>
+      <groupId>com.wordnik</groupId>
+      <artifactId>swagger-annotations</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
```
jpkrohling: Does it work, even without explicit swagger annotations?

lucasponce:
Cleaned the pom.xml with suggestions and adding the commit on the splash info.

jpkrohling:
LGTM

PR #48, opened 2016-08-05
Hawkular-1107 Add more event types

jshaughn:
Enhance the CommandEventListener to handle a Undeploy responses
as well as datasource responses. Note that DS messages are a future,
so support is written in anticipation.


pilhuhn:
Thanks @jshaughn This is good to have it. To be able to work with it in ManageIQ

jshaughn:
@pilhuhn The Undeploy events should start working when PR #10287 gets merged.  The Datasource events will hopefully "just work" when we start utilizing those requests from MIQ.

PR #50, opened 2016-08-09
Allow to specify the TTL for metrics data for Docker use.

pilhuhn:
Default is 14 days.

@pgier would you please check?


pgier:
The Docker image builds fine for me with the changes.  I verified that the new var gets passed to the generated Dockerfile, but I'm not sure how to actually test that the TTL setting takes effect in the running image.

pilhuhn:
@pgier I saw the new settings when retrieving metrics with HawkFx and verbose logging on.

PR #53, opened 2016-08-12
Minor build fixes

pgier:


docker-dist/README.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+By default, this will create a docker image with the name "<username>/hawkular-services".
+
+  docker images
+
+== Running Hawkular Services
+
+Next, start up the hawkular services container and point it to the Cassandra instance.
+
+  docker run -e CASSANDRA_NODES=<cassandra ip> <username>/hawkular-services
```
pilhuhn: I think one should use docker-linking in this case. As described in http://pilhuhn.blogspot.de/2016/06/using-hawkular-services-via-docker.html "Running without docker-compose"

PR #54, opened 2016-08-18
Trivial: add miqResourceType param to addEvent

jshaughn:


jmazzitelli:
beautiful work! programming that will withstand the test of time and be studied by CS students for generations to come.

PR #57, opened 2016-08-22
Use $HOSTNAME instead of `hostname` command

pgier:
The hostname command is not always available, specifically in a RHEL 7 container


theute:
Travis doesn't seem happy, but that's unrelated to this PR.
LGTM

PR #58, opened 2016-08-24
HAWKULAR-1112 - Increased number of retries on itests

jpkrohling:
@theute , @jmazzitelli , could one of you please review this one? If this passes on Torii as well, I'll make a Services release afterwards.


jmazzitelli:
For the record, in my tests, I changed the delay time (from 100 -> 1000) not the retry count from 50 to 500 as this PR does. But since this is green, I'll assume its OK to fire requests to h-metrics every 100 milliseconds up to 500 times :)

PR #72, opened 2016-10-06
Fixing github call for bot.

josejulio:
~~DO NOT MERGE (yet)~~


josejulio:
Is OK to review and merge now.

pilhuhn:
Are you sure that this Heroku thingy is reliable? What happens when it is down?

josejulio:
It won't report the gem status if is down. (e.g. no @hawkular-bot message).
If needed, I could change from heroku to somewhere else.

josejulio:
As for if its reliable. I can't guarantee that it won't be down, as I'm using a free account.

PR #76, opened 2016-10-18
hawkular-1127 fix NPE in backfill

jshaughn:
- the root cause was hwkmetrics-512, the MetricsService was not bound in JNDI
- also improve the lazy init of services


jshaughn:
@pilhuhn Do Not Merge : Still need to replace Metrics snapshot dep, but the other (minimal) changes are free to review.  Hope to have a SRC dep later today for metrics.

Issue #77, opened 2016-10-20
Admin-Token for metrics isn't set on docker image


garethahealy:
The latest metrics introduces "-Dhawkular.metrics.admin-token". As its not on the base docker image, any calls to /tenants fail.

https://github.com/hawkular/hawkular-metrics/blob/b5e1a337e206c961c3b9e8b47d40e28d081c2c59/api/metrics-api-jaxrs/src/main/java/org/hawkular/metrics/api/jaxrs/filter/AdminFilter.java#L67

https://github.com/hawkular/hawkular-services/blob/master/docker-dist/src/main/resources/startcmd.sh

pilhuhn:
What is the use case / need for this?

garethahealy:
As per the comment on the PR, i dont think its needed anymore. I didn't realise you could use the ADMIN_TOKEN env var before raising the PR.

Happy to close if you dont think its needed as well.

PR #78, opened 2016-10-20
Added admin token to startcmd

garethahealy:
https://github.com/hawkular/hawkular-services/issues/77


garethahealy:
Not sure if this is actually needed, as you can just use the env var ADMIN_TOKEN

garethahealy:
Not sure if this is actually needed, as you can just use the env var ADMIN_TOKEN

PR #79, opened 2016-10-25
Bump metrics and alerts

pilhuhn:


jshaughn:
I think metrics should be 0.21.2.Final.

PR #87, opened 2016-11-03
HAWKULAR-1132 Parse response from status endpoint to know the actual …

josejulio:
…status of hawkular-services, metrics, alerts and inventory.

https://bugzilla.redhat.com/show_bug.cgi?id=1381616


josejulio:
Wrong Jira

PR #104, opened 2016-11-29
Replace srcdeps-maven-plugin with srcdeps core extension

ppalaga:


ppalaga:
@pilhuhn or whoever else, could you please review and eventually merge?

ppalaga:
@pilhuhn not responding, could @Jiri-Kremser review maybe?

.travis.yml
```diff
@@ -24,15 +25,12 @@ addons:
     - oracle-java8-installer
 
 install:
-- bash .travis.install.maven.sh "3.3.9" "${HOME}/mvn-home"
```
pilhuhn: Why did you remove those + the associated install script?

ppalaga: Because I replaced it with `mvnw` - that's a standard way to do the same thing (download and install some specific Maven version). Moreover `mvnw` can be used by both CI and developers.

.travis.hawkular-client.ruby.sh
```diff
@@ -26,9 +26,9 @@ fi
 export ADMIN_TOKEN=4f4a1434-8cb3-11e6-ae22-56b6b6499611
 
 # Create docker image.
-mvn -s .travis.maven.settings.xml -Pdev,dozip clean install
+./mvnw -s .travis.maven.settings.xml -Pdev,dozip clean install
 pushd docker-dist
-mvn -s ../.travis.maven.settings.xml docker:build
+./mvnw -s ../.travis.maven.settings.xml docker:build
```
josejulio: I was wondering why there isn't a hawkular status on the ruby-client here.
After checking the travis log i found the following at the bottom:
```bash
+pushd docker-dist
~/build/hawkular/hawkular-services/docker-dist ~/build/hawkular/hawkular-services
+./mvnw -s ../.travis.maven.settings.xml docker:build
.travis.hawkular-client.ruby.sh: line 31: ./mvnw: No such file or directory
```

I guess this one should be `../../mvnw`, notice the pushd.

Jiri-Kremser: Can I build the project with normal maven, or is everyone forced to use this `mwnw` wrapper or is it only needed if for the src-dep to work?

If it's the only way how to build the project, what will happen if the src-dep dependency is somewhere in the `pom.xml` and normal `mvn` command is used? Or how to build the nested sub-modules? 
`./../../../../../mvnw clean install` ? Or alter the `$PATH` ? Both methods are cumbersome :/

ppalaga: > I guess this one should be `./../mvnw`, because of the pushd. 

Good catch, @josejulio . Let me fix it.

ppalaga: > Can I build the project with normal maven?

Yes, as long as your maven version is the same or close enough as the one in `.mvn/wrapper/maven-wrapper.properties` ATM, it has to be Maven 3.3.1+.

ppalaga:
bf4cc5b replaces 49baab1 fixing the `mvnw` path in https://github.com/hawkular/hawkular-services/pull/104/commits/ddd0394a49cd6200c566926e6d67ee15534fe7d4#diff-5922539fc070ae984a1f0132c80e2e17R31

Is this good to merge now?

ppalaga:
> Oh, I wasn't aware we need to install our special maven in travis.
>
> Hmm, if |mvn -N io.takari:maven:wrapper| downloads the |mvnw| shell
> script, so why don't we just have something like this:
>
> #!/bin/bash
> [ -f ./mvnw ] || mvn -N io.takari:maven:wrapper
> ./mvnw $@
>
> in a file called |build.sh|.

Wrapping a wrapper :D !!! Well, to run the above script or `mvn -N 
io.takari:maven:wrapper` you need mvn installed, while you do not need 
to have `mvn` installed to run `mvnw`. I think that's important in many 
cases, esp. in CI. Second, your script will run on unix like systems 
only. There is `mvnw.cmd` for Windows OOTB.

Jiri-Kremser:
ok, you convinced me. LGTM

PR #107, opened 2016-12-14
HAWKULAR-1138 move to okhttp 3.x

jmazzitelli:

PR #108, opened 2016-12-16
bump inventory to 1.1.3.Final

metlos:


metlos:
The build fails during xml transformations while building feature pack because of no output in 10 minutes. Not sure this is caused by inventory.

Could someone with enough privs restart the builds, pls?

jmazzitelli:
for the record, this is going to be incorporated when PR #107 is merged. That PR requires the new commons and will also pull in the new inventory along with it.

Of course, this required PR #104 to be merged first.

jmazzitelli:
This PR can be closed since this inventory release had been incorporated in another PR - see PR #107

PR #111, opened 2017-01-25
HAWKULAR-1188 Event on Server addition or removal

jshaughn:
Listen for Inventory events creating or removing servers. Note there
are several server resource types, WildFly Server, Host Controller,
etc...

Refactored a bit now that we have different sorts of inventory
events generating server-side work and multiple listeners generating
Events for MIQ consumption.


jshaughn:
@pilhuhn The screenshot below shows what I have here.   I need to know if this satisfies the requirement.  Because these are not representing an alert or failure I made these *detail* level events (not summary), so they only show if you click the checkbox. There are links to the provider and to the server resource.

hawkular-1188

jshaughn:
@pilhuhn Note, To do this I used the generic "hawkular_event" that I already have defined in MIQ.  So, there is no MIQ side work/coordination involved.

jshaughn:
@pilhuhn Nudge...

pilhuhn:
LGTM - especially that those magically show up in MiQ

PR #116, opened 2017-02-08
Provide an explicit Dockerfile and modified standalone.conf.

pilhuhn:
Also chown the dirs to jboss.

This later step may not be needed if the base image already does this.

### DIFF
--- a/docker-dist/pom.xml
+++ b/docker-dist/pom.xml
@@ -119,33 +119,15 @@
               <alias>hawkular-services</alias>
               <name>${user.name}/hawkular-services</name>
               <build>
-                <from>${docker.base-image}</from>
-                <maintainer>Hawkular Team, hawkular-dev@lists.jboss.org</maintainer>
+                <dockerFile>Dockerfile</dockerFile>
+                <args> <!-- TODO really needed or do we just want to modify the Dockerfile -->
+                  <agent_enable>${hawkular.agent.enabled}</agent_enable>
+                  <metrics_ttl>${hawkular.metrics.ttl}</metrics_ttl>
+                </args>
                 <tags>
                   <tag>${docker.tag.prefix}latest</tag>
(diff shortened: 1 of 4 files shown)


Jiri-Kremser:
When Openshift runs the container, it doesn't take into consideration the user that is defined in the dockerfile, but it uses 'random' UID. The UID range can be set per project by setting the openshift.io/sa.scc.uid-range metadata annotation, but by default we can assume it's pretty much random.

So chowning everything to jboss user doesn't help us.

this is what the OS online course says:

In a pure Docker environment, simple Dockerfiles may generate images that fail to run in OSE 3. Most of the time, this is because the image needs to run as root or some other fixed UID. But OSE runs containers using a random UID. The fix is making sure folders written to while running the container have a+rwx permissions.

pgier:
I like having the explicit Dockerfile vs. the auto-generated one.  One negative though  is that we can no longer parameterize the base image name and other info.  I submitted a PR to the docker-maven-plugin (https://github.com/fabric8io/docker-maven-plugin/pull/686) to allow variable substitution in the Dockerfile, so hopefully with a future release of the plugin we can more easily switch out the base image.

Jiri-Kremser:
lgtm, the travis went crazy, but I was able to build it locally and use it in openshift with the `openshift/origin-metrics-cassandra:v1.4.1` as C* image and it works like a charm

pilhuhn:
I think this bombs on travis as the used `docker-compose.yml` file has no volume defined for the data directory.
```
volumes:
    - /tmp/opt/hawkular/server:/opt/data
```
(But that is more or less a guess)

pilhuhn:
Travis is now (almost) happy after the volume has been added. There are still 2 failures where operations fail, but inside container they would not be expected to work anyway.

Merging this now.

PR #118, opened 2017-02-10
Configure travis to push to docker hub for each tag

pgier:


rhqci:

pilhuhn:
Shouldn't we do that from Jenkins and only when a release is done?

pilhuhn:
Mvn is unhappy with bad/missing license in `.travis.docker-push.sh`

pgier:
Updated the Jenkins configuration to do the push, so this PR is no longer needed.

PR #122, opened 2017-02-15
Set the immutable flag explicitly. Also set feedid to hostname in container

pilhuhn:


docker-dist/src/main/resources/standalone.conf
```diff
@@ -74,6 +74,7 @@ fi
 
 # How set hawkular-agent specific things
 JAVA_OPTS="$JAVA_OPTS -Dhawkular.agent.machine.id=`hostname` -Djboss.server.name=${HOSTNAME} -Dhawkular.agent.in-container=true"
+JAVA_OPTS="$JAVA_OPTS -Dhawkular.rest.feedId=`hostname` -Dhawkular.agent.immutable=true"
```
pgier: Can ${HOSTNAME} be used instead of `hostname`?  The "hostname" command may not be available on all containers.

pilhuhn:
@pgier Ok with you now?

PR #132, opened 2017-03-29
upgrade agent

jmazzitelli:
move to new agent that let's you set a container id


rhqci:

jmazzitelli:
@pilhuhn I just ran a quick smoke test and this is working. I started the h-services server with -Dhawkular.agent.container.id=mazz-container-id and i see the platform resource now has a container ID with that value - it is next to the already existing machine ID property. So this is working as expected.

Note you can alternatively add a "container-id" property in the <platform> configuration element in the configuration file (either in standalone.xml or the YAML file for a java agent). But I suspect when running a container image within an OpenShift  pod, you will be passing in the -Dhawkular.agent.container.id option.

pilhuhn:
Thanks @jmazzitelli

PR #136, opened 2017-04-04
HAWKULAR-1208 Replace Server-Added Event Notification

jshaughn:
- add new /notification endpoint for rest-based notifications
   - Migrate PingHandler -> ApiHandler for all /api services
- add a few ResponseUtil utils for easy message-based responses
- remove inventory event handler for add/remove server events
- ListenerUtils: add ability to generate event only if it doesn't exist
- comment out unused hystrix filters and dependencies
- replace wordnik.swagger deps with io.swagger deps

@jmazzitelli @pilhuhn So this PR is a POC for server-side support of the "Server Added" event feature.   It's basically a simple service at the services [glue] level that can take a notification (type and property map) and do what we want with it.  For a RESOURCE_ADDED notification it would generate an event for MIQ consumption, *if necessary*.  If the event [id] already exists it won't create another one.

I think by locating the logic in the glue code and just letting the agent generate notifications based on some new resource-type config meta-data (see dev list discussion), we may be in decent shape.  The agent doesn't have to be smart or change anything based on a change to MIQ integration or whatever. We eliminate the inventory hooks and don't rely on metrics.  The glue code just interacts with the events api offered by alerting.

If this looks promising we could look at adding the agent support.


FilipB:
retest this please

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/ListenerUtils.java
```diff
@@ -47,25 +47,35 @@ public ListenerUtils() {
...
+            String miqEventType,
             String miqResourceType, String miqMessage) {
         try {
             init();
 
             String tenantId = resourcePath.ids().getTenantId();
-            String eventId = UUID.randomUUID().toString();
+            eventId = (null == eventId || eventId.isEmpty()) ? UUID.randomUUID().toString() : eventId;
+
+            if (checkExists) {
```
lucasponce: I am out of context and probably my question needs more info.
What is the use case of checkExists ?
I mean, when adding the event, if exists it returns an error code with message that perhaps can be used for this.
(It's a minor optimization).

jshaughn: No, you're in context.  CheckExists allows the caller to be explicit about his intent.  If checkExists is true the call is successful regardless of whether the event exists.  If checkExists is false and the event exists the call is a failure, (currently just a log message but I may change this to return a boolean).  As for the impl, you're right I could have deferred the check to the alerting call but then I would have had to parse our the reason of the 400, this is just easier and more clear.  Thanks for looking at the PR!

hawkular-rest/hawkular-rest-api/src/main/java/org/hawkular/rest/ApiHandler.java
```diff
@@ -0,0 +1,149 @@
...
+            "Host Controller",
+            "WildFly Server"));
+
+    private void handleResourceAdded(Notification notification) {
+        String resourceType = notification.getProperties().get("resourceType");
+        if (isEmpty(resourceType)) {
+            throw new IllegalArgumentException("Required Property [resourceType] is missing or is an invalid type.");
+        }
+        String resourcePath = notification.getProperties().get("resourcePath");
+        CanonicalPath cp;
```
jotak: Can you remove the dependency on CanonicalPath? Since the hawkular-inventory dependency will be removed, could we just use its string representation?

In the wildfly agent, there's still a usage of inventory's data model so I copied related classes to the agent's repo.

If some of the data model is still needed in hawkular-services and cannot be worked out, then maybe we should rather put some stuff in common.

jshaughn: This is something to discuss, maybe with @jmazzitelli as well.  As far as I know we are still using canonical paths in resource names and such, and CanonicalPath is part of commons, not inventory.  It's quite useful to keep CanonicalPath as it has all of the built in parsing for pulling out the tenant, resource type, etc.  If we're going to still be using CP strings in what the agent sends us, I think I'd prefer to keep using this on the server.  Otherwise I think we'd need to re-build the same parsing abilities.

jotak: Oh you're right, this one is in common.
But the problem is still there for classes like "org.hawkular.inventory.api.model.StructuredData" and the like.

jshaughn: Certainly you're right, we need to remove all use of inventory API that no longer exists.  This PR doesn't introduce use of StructureData, but it also does not finish removing all of InventoryEventListener. That will still need to happen.

jshaughn:
@jmazzitelli @jotak @pilhuhn  I think this can be merged.  The agent side PR is being generated but will be marked DoNoMerge as is currently depends on a SNAPSHOT version of the server (and also needs some test).  Step 1 is to merge this PR, I think, assuming no objections.

jshaughn:
Setting back to DNM after a chat with @jmazzitelli.  Will move new client jar module to commons to avoid a circular dep issue with the agent...

jshaughn:
Updated to use new common hawkular-client-api.jar to get rid of circular dep issues with the agent.  Still set to DNM because of snapshot deps on commons and the agent.

jmazzitelli:
Running all of this results in the following for me:

```
14:55:59,806 WARN  [org.hawkular.rest] (default task-19) RestEasy exception, : javax.ws.rs.NotFoundException: RESTEASY003210: Could not find resource for full path: http://localhost.localdomain:8080/hawkular/api/notification
...
14:55:59,812 ERROR [org.hawkular.agent.monitor.storage.HawkularStorageAdapter] (OkHttp http://localhost.localdomain:8080/...) HAWKMONITOR010083: Failed to store notification: {"type":"RESOURCE_ADDED","properties":{"resourcePath":"/t;hawkular/f;d9cd8940-60e4-4944-b0dd-3342a1a7d6d2/r;Local~~","resourceType":"WildFly Server"}}: java.io.IOException: status-code=[500], reason=[Internal Server Error], url=[http://localhost.localdomain:8080/hawkular/api/notification]
	at org.hawkular.agent.monitor.storage.HawkularStorageAdapter$4.onResponse(HawkularStorageAdapter.java:513)
```

jshaughn:
retest this please

jshaughn:
retest this please

jshaughn:
retest this please

PR #142, opened 2017-04-11
Replace hawkular-wildfly-agent with hawkular-javaagent

josejulio:
~~Still need to test the docker-dist and port the SSL configuration from hawkular-wildfly-agent to hawkular-javaagent.~~


rhqci:

josejulio:
Hawkular javaagent `local-dmr` when outside of Wildfly (i.e. when not using hawkular wildfly agent) is a `remote-dmr` that tries to guess its connection values with wildfly environment variables.
Because is not really configurable, had to disable it (`local-dmr`) and using a `remote-dmr` instead.

I couldn't found any tool like `xslt` to edit the javaagent `YAML` (configuration file) so I ended using a groovy script (maven side) as part of the build.

Also now http is turned off on the management interface and turned on https.

josejulio:
Still need to update the itests to actually use the javaagent.

josejulio:
Some client tests are failing because the DMR endpoint is now called "Hawkular Services DMR". I'm setting it to local. Other tests are failing because some tests use subsystem=hawkular-wildfly-agent which no longer exists.

@Jiri-Kremser could you please review the changes? I had to port some SSL configuration from XML to YAML

feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template.xsl
```diff
@@ -22,31 +22,11 @@
 
   <xsl:output method="xml" version="1.0" encoding="UTF-8" indent="yes" xalan:indent-amount="4" standalone="no" />
 
-  <xsl:param name="hawkular.agent.enabled" select="'true'"/>
-  <xsl:param name="hawkular.rest.user" select="''"/>
-  <xsl:param name="hawkular.rest.password" select="''"/>
   <xsl:param name="hawkular.embeddedc" select="'false'"/>
 
-  <!-- Add the default user and password if they were passed in through the parameters -->
+  <!-- Add the embedded cassandra if it was passed in through the parameters -->
```
Jiri-Kremser: I am wondering if the embedded C* is still working, I haven't used it for more that a year

feature-pack/src/main/script/build-javaagent-config.groovy
```diff
@@ -0,0 +1,67 @@
...
+config['managed-servers']['remote-dmr'][0]['port'] = template('jboss.management.http.port', 9990)
+config['managed-servers']['remote-dmr'][0]['username'] = template('hawkular.rest.user')
+config['managed-servers']['remote-dmr'][0]['password'] = template('hawkular.rest.password')
+config['managed-servers']['remote-dmr'][0]['resource-type-sets'].add('Hawkular')
+
+parser.dump(config, new FileWriter(JAVAAGENT_CONFIG))
+
+// Create a SSL version
+config['storage-adapter']['security-realm'] = 'HawkularAgentRealm'
+config['storage-adapter']['url'] = template('hawkular.rest.host', 'https://127.0.0.1:8443')
```
Jiri-Kremser: This does the same thing as line 39 so you are overwriting the value, is this intentional?

feature-pack/src/main/script/build-javaagent-config.groovy
```diff
@@ -0,0 +1,67 @@
...
+    "\${${property}:${defaultValue == null ? properties[property] : defaultValue}}".toString()}
+
+// Configure javaagent for non ssl mode
+config['subsystem']['enabled'] = template('hawkular.agent.enabled')
+config['storage-adapter']['username'] = template('hawkular.rest.user')
+config['storage-adapter']['password'] = template('hawkular.rest.password')
+config['storage-adapter']['url'] = template('hawkular.rest.host', 'http://127.0.0.1:8080')
+
+// Disable local-dmr and use a remote-dmr which allows a more detailed configuration.
+config['managed-servers']['local-dmr']['enabled'] = false
```
josejulio: Because javaagent is outside wildfly we can't really use a `local-dmr`.
On the javaagent, `local-dmr` is just a `remote-dmr` pointing to a local IP. The agent tries to guess the correct values, a remote-dmr should be used [1] if these don't fit the use case.

[1] https://github.com/hawkular/hawkular-agent/blob/master/hawkular-javaagent/src/main/java/org/hawkular/agent/javaagent/config/ConfigConverter.java#L280-L283

feature-pack/src/main/script/build-javaagent-config.groovy
josejulio: Yes, is intentional.

We are creating two files on l[ine 50](https://github.com/hawkular/hawkular-services/pull/142/files/9ec799942605df0c4fc17b48c922ab63e6bc418b#diff-8e275e601705501f5c60fd646fa25c48R50) we dump the current configuration to `hawkular-javaagent-config.yaml`, whatever is after it is saved to `hawkular-javaagent-config-ssl.yaml` on [line 67](https://github.com/hawkular/hawkular-services/pull/142/files/9ec799942605df0c4fc17b48c922ab63e6bc418b#diff-8e275e601705501f5c60fd646fa25c48R67)

feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template.xsl
josejulio: I tested it outside docker, it still works. 
I suppose one should set the cassandra host to `localhost` for it to work on the docker images.

jmazzitelli: embedded C still works - I use it all the time. "mvn clean install -Pembeddedc,dev" still works when you run that and run the server it builds

feature-pack/src/main/script/build-javaagent-config.groovy
Jiri-Kremser: :+1: I missed the `parser.dump` command

feature-pack/src/main/script/build-javaagent-config.groovy
jmazzitelli: it is ok to use local-dmr if we can find out where the management endpoint is bound. I would prefer to use local-dmr if that is what you were using before, only because it means less changes on the client side. But you can use whatever is easiest for you to use.

josejulio: I couldn't find any way to specify we are going to use SSL without messing with more agent code. It currently uses http-remoting on `local-dmr` [1].

[1] https://github.com/hawkular/hawkular-agent/blob/master/hawkular-javaagent/src/main/java/org/hawkular/agent/javaagent/config/ConfigConverter.java#L303

feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template.xsl
Jiri-Kremser: Apart from that it's not good to have 1 big image with everything, I doubt we ship the C* in the h-services docker image. iirc it's there only if the mvn profile / property is used

josejulio: That's correct, as @jmazzitelli pointed out. Cassandra will only be embedded if -Pembeddedc is used.

josejulio:
Rebased and now one test seems to be failing. I'll take a look at it.
```
Caused by: java.lang.AssertionError: [Request{method=GET, url=http://127.0.0.1:8080/hawkular/alerts/events?tagQuery=miq.event_type%20%3D%20hawkular_event%20AND%20miq.resource_type%20%3D%20MiddlewareServer, tag=null}] should have returned a json array with size == 1, while it returned [[]] expected [true] but found [false]
```

pilhuhn:
Please wait with merging this until we have sorted out the Inventory.next pieces.

josejulio:
Tests seems to be failing due to a bug on the javaagent not loading the NotificationTypes. 
This should be fixed when a new agent with https://github.com/hawkular/hawkular-agent/pull/323 is released.
When that happens I'll rebase (also forgot to delete one [unused file](https://github.com/hawkular/hawkular-services/blob/master/itest/src/test/xsl/standalone.xsl)) and push.

jmazzitelli:
PR https://github.com/hawkular/hawkular-agent/pull/323 has been merged

josejulio:
thanks @jmazzitelli 

I'll use that version to check if travis is happy with that.

feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template.xsl
Jiri-Kremser: right, his comment was faster than mine ;)

josejulio:
Rebased this one to have it green.

PR #145, opened 2017-04-24
HAWKULAR-1210 Remove legacy REST API

jshaughn:
- Simplify rest war by removing unused services and associated support
- Refactor CORS support with commons-based approach


jshaughn:
@Jiri-Kremser @pilhuhn , This PR cleans up unused rest endpoints, and their deps, from the services rest war.  Please review and let me know if I was too aggressive, or if there is anything I missed.   I think it would be useful to get this merged and and then rebase the inv-on-metrics branch against the changes.  Thanks, Jay.

jshaughn:
retest this please

Jiri-Kremser:
lgtm, hopefully we wont need the Hystrix in the future, and if yes, it's probably better to put it directly to the metrics to save 1 http call (because I was calling the component APIs via REST)

perf tests are red again, but it's not caused by this PR

(I can't merge, (no rights))

jshaughn:
@Jiri-Kremser Thanks for the review, I'll merge based on your feedback.

PR #146, opened 2017-04-24
Container improvements

pgier:
Just some minor improvements to the Docker and openshift stuff


pgier:
I chose /var/opt/hawuklar for now instead of /var/lib/hawkular since this is what FHS recommends when the binaries are installed to /opt.  If we end up packaging Hawkular in an RPM in the future, then the data dir could be moved to /var/lib.

PR #148, opened 2017-04-25
HAWKULAR-1211 remove inventory from splash page

jshaughn:


josejulio:
The client tests got stuck waiting by a now gone inventory. This made the test to timeout and fail (even though it already passed).
Latest client code won't wait for inventory, A restart should make this one go green.

PR #150, opened 2017-04-27
HAWKULAR-1212 Fix backfill

jshaughn:
- remove lookup of inventory service
- change avail metric filtering from using the metric type properties
  (which are empty) to its assigned MetricDataType
- add Base64 decoding when stitching together the chunks


jshaughn:
@jotak, thanks for the help and please review.

PR #152, opened 2017-05-02
Upgrade hawkular metrics 0.26.0 -> 0.26.1

pgier:


pilhuhn:
Ruby client checks fail because with the javaagent the inventory structure has changed and thus subsystem=agent can't be found. Same with the in-container/immutable flags.

PR #154, opened 2017-05-02
Remove unnecessary wf-agent stuff now that javaagent is in use.

jshaughn:
Now that javaagent has been merged, I'm fairly sure this stuff can also be removed.  @jmazzitelli please review.


jshaughn:
I should have mentioned that things seemed to work fine after these removals, with the -Pdev build.

jmazzitelli:
What this will do is effectively disallow anyone to install the agent as a subsystem in any EAP. This is because the download war (and the installer and extension zips that it allowed people download) will no longer be available in the server.

If this is what we want (that is, no longer want anyone to use the agent as a subsystem) then we should also remove all the agent subsystem code (and related code, like the installer) from the agent github repo. Though I am unsure we are ready to make that commitment.

jshaughn:
@jmazzitelli I'm fine if we leave it, I saw what looked like wf-agent code hanging around and wasn't sure we needed it, but maybe we want to preserve the option of injecting an agent as a subsystem.  I'm not 100% sure why though (eap6 support?), because it seems like we now have 3 agents that do basically the same thing:  wf-agent, wf-swarm-agent, javaagent.  Maybe @pilhuhn has an opinion.

josejulio:
I left that there because AFAIK there isn't any other option to run the agent in domain mode.

jmazzitelli:
We cannot merge this PR. We need the download war and installer jar to support the agent subsystem when running in WildFly in domain mode.

PR #157, opened 2017-05-04
Inventory sanity checks

jotak:
- Adding unit tests on inventory
- Adding sanity checks on chunks reading; mazz had a case where data was not being read correctly, I'm not actually sure if data was somehow corrupted or if it was race condition when writing to C* and reading at the same time, thus reading incomplete data.

@jshaughn I've added some log.warn here, when such sanity check fails. Is "warn" the correct level to put here? Hopefully we shouldn't see it regularly; or if we do then we have a problem.

### DIFF
--- a/hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/InventoryHelper.java
+++ b/hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/InventoryHelper.java
@@ -24,7 +24,6 @@ import java.util.Base64;
 import java.util.Base64.Decoder;
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.zip.GZIPInputStream;
@@ -34,6 +33,7 @@ import org.hawkular.inventory.api.model.Metric;
 import org.hawkular.inventory.api.model.MetricType;
 import org.hawkular.inventory.json.InventoryJacksonConfig;
 import org.hawkular.inventory.paths.RelativePath;
+import org.hawkular.listener.exception.InvalidInventoryChunksException;
 import org.hawkular.metrics.core.service.MetricsService;
 import org.hawkular.metrics.core.service.Order;
 import org.hawkular.metrics.model.DataPoint;
@@ -97,14 +97,14 @@ public final class InventoryHelper {
                     return metricsService.findStringData(metric.getMetricId(), 0, currentTime,
                             false, 0, Order.DESC)
                             .toList()
-                            .map(InventoryHelper::rebuildFromChunks)
-                            .doOnNext(inv -> {
-                                if (inv == null) {
-                                    LOG.warnf("Could not rebuild inventory from chunks for metric type %s. " +
-                                            "Expecting future calls to work correctly", metric.getId());
+                            .map(dataPoints -> {
+                                try {
+                                    return rebuildFromChunks(dataPoints);
+                                } catch (InvalidInventoryChunksException e) {
+                                    e.addContext("metricId", metric.getId());
+                                    throw Throwables.propagate(e);
                                 }
                             })
-                            .filter(Objects::nonNull)
                             .map(inv -> inv.getStructure().get(RelativePath.fromString("")))
                             .filter(bp -> bp instanceof MetricType.Blueprint)
                             .map(bp -> (MetricType.Blueprint) bp);
@@ -136,11 +136,13 @@ public final class InventoryHelper {
                     return metricsService.findStringData(metric.getMetricId(), 0, currentTime,
                             false, 0, Order.DESC)
                             .toList()
-                            .map(InventoryHelper::rebuildFromChunks)
-                            .doOnNext(inv -> {
-                                if (inv == null) {
-                                    LOG.warnf("Could not rebuild inventory from chunks for metrics of type" +
-                                            " %s. Expecting future calls to work correctly", metricType.getName());
+                            .map(dataPoints -> {
+                                try {
+                                    return rebuildFromChunks(dataPoints);
+                                } catch (InvalidInventoryChunksException e) {
+                                    e.addContext("metricType", metricType.getId());
+                                    e.addContext("metricId", metric.getId());
+                                    throw Throwables.propagate(e);
                                 }
                             })
                             .map(inv -> extractMetricsForType(inv, metricType.getId()))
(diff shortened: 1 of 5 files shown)


rhqci:

pilhuhn:
>  I've added some log.warn here, when such sanity check fails.

Shouldn't we fail the test then? If the test passes and is green, no one will look at the logs.

jotak:
1/ This is not for test, this is in production code.
2/ Maybe you're right, but currently I lack information on why it happened once. I'd like to know if it's something recoverable automatically on next backfill pass, or not.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/InventoryHelper.java
```diff
@@ -78,13 +81,29 @@ private InventoryHelper() {
...
                 .flatMap(metric -> {
-                    return metricsService.findStringData(metric.getMetricId(), 0, System.currentTimeMillis(),
+                    return metricsService.findStringData(metric.getMetricId(), 0, currentTime,
                             false, 0, Order.DESC)
                             .toList()
                             .map(InventoryHelper::rebuildFromChunks)
+                            .doOnNext(inv -> {
+                                if (inv == null) {
+                                    LOG.warnf("Could not rebuild inventory from chunks for metric type %s. " +
+                                            "Expecting next backfill job to work correctly", metric.getId());
```
jshaughn: @jotak , I think warn-level is fine, I think.  We want to know if it's failing consistently.  It is a little odd that the message contains knowledge of the caller (backfill job) but probably not worth changing.

jotak:
To give more material on the reason of that sanity check: mazz sent me that stacktrace he saw once (and only once):

FAILED: testRemoveXaDatasource
java.lang.ArrayIndexOutOfBoundsException
        at java.lang.System.arraycopy(Native Method)
        at org.hawkular.agent.itest.util.ITestHelper.rebuildFromChunks(ITestHelper.java:124)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
        at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:747)
        at java.util.stream.ReduceOps$ReduceTask.doLeaf(ReduceOps.java:721)
        at java.util.stream.AbstractTask.compute(AbstractTask.java:316)
        at java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:731)
        at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
        at java.util.concurrent.ForkJoinPool.helpComplete(ForkJoinPool.java:1870)
        at java.util.concurrent.ForkJoinPool.externalHelpComplete(ForkJoinPool.java:2467)
        at java.util.concurrent.ForkJoinTask.externalAwaitDone(ForkJoinTask.java:324)
        at java.util.concurrent.ForkJoinTask.doInvoke(ForkJoinTask.java:405)
        at java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:734)
        at java.util.stream.ReduceOps$ReduceOp.evaluateParallel(ReduceOps.java:714)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
        at org.hawkular.agent.itest.util.ITestHelper.extractStructuresFromResponse(ITestHelper.java:97)
        at org.hawkular.agent.itest.util.ITestHelper.getBlueprintsByType(ITestHelper.java:209)
        at org.hawkular.agent.itest.util.ITestHelper.waitForNoResourceContaining(ITestHelper.java:285)
        at org.hawkular.agent.ws.test.DatasourceCommandITest.testRemoveXaDatasource(DatasourceCommandITest.java:361)

This is in the java agent integration tests, but if it happens there, it could probably happen here also.
The error in System.arrayCopy from "rebuildFromChunks" means that, somehow, some expected datapoints were not present (at least this is the more evident cause that I see).

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/InventoryHelper.java
```diff
@@ -139,12 +176,20 @@ private static ExtendedInventoryStructure rebuildFromChunks(List<DataPoint<Strin
                 if (master.length == 0) {
                     return null;
                 }
+                if (nbChunks > datapoints.size()) {
+                    // Sanity check: missing chunks?
+                    return null;
```
jshaughn: optional: this may deserve a log.warn as well, because we won't know the reason from the log message generated from the caller.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/InventoryHelper.java
jotak: Yes, I know, but it's to help a little bit whoever reads this message: to let him know that he shouldn't see this message constantly.

jotak:
@jshaughn , what do you think about @pilhuhn remark? Better to let it fail, or to log & continue?

jotak:
added suggested warnings & squashed

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/InventoryHelper.java
```diff
@@ -78,13 +81,29 @@ private InventoryHelper() {
...
                             false, 0, Order.DESC)
                             .toList()
                             .map(InventoryHelper::rebuildFromChunks)
+                            .doOnNext(inv -> {
+                                if (inv == null) {
+                                    LOG.warnf("Could not rebuild inventory from chunks for metric type %s. " +
+                                            "Expecting future calls to work correctly", metric.getId());
+                                }
+                            })
                             .filter(Objects::nonNull)
```
jshaughn: I could go either way.  The way it is you could succeed in doing some backfilling because some rebuilds may succeed .  But I also don't mind if this call chain actually throws an Exception (likely rooted down below in RebuildFromChunks), because well, from the perspective of this call, it's a bad/unexpected situation.  If we threw an exception instead, we could defer the catching/handling to the caller.  In which case we may end up doing the same thing, logging from the backfiller, but it may be a cleaner approach.

pilhuhn:
Is this ready to be merged?

jotak:
@pilhuhn no yet ; I'm trying to do as Jay suggests

jotak:
@jshaughn / @pilhuhn  can you look at this version? (w/ exceptions thrown)

jshaughn:
@jotak, this looks good, we could merge as-is (after fixing the license header) or we could also add handling in the backfiller for the new exception, letting it continue to backfill anything it can do successfully.  I don't mind if that work is done as a follow-up.

jotak:
Ok, so it's ready to merge.
I must see how to "delay" error processing with rx. It will be for another PR.

jshaughn:
@pilhuhn I'll let you merge if you feel the DNM label can be removed.

pilhuhn:
@jshaughn  @jotak could already have removed the label. I only put it so that others also have an indicator that Joel wants to make some changes

PR #165, opened 2017-06-02
[BZ-1455877] Resolve real path for cacerts before setting permissions

josejulio:
`$JAVA_HOME/jre/lib/security/cacerts` can be a symlink, thus we need to resolve the real path to change the permissions to that file.


rhqci:

josejulio:
@Jiri-Kremser 
Could you please review?

Jiri-Kremser:
lgtm

PR #167, opened 2017-06-28
Provide docker-compose file at project root

jotak:
I think it's always a good point to provide a docker-compose file to quickly start any software :)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,37 +15,20 @@
 # limitations under the License.
 #
 
-# Example of an instrumented wildfly that talks to hawkular
-#hawkfly:
-#  image: "pilhuhn/hawkfly:latest"
-#  ports:
-#    - "8081:8080"
-#  links:
-#    - hawkular
-
-
-# The hawkular-server
 hawkular:
   image: "hawkular/hawkular-services:latest"
   ports:
     - "8080:8080"
     - "8443:8443"
-    - "8787:8787"
-  volumes:
-    - /tmp/opt/hawkular/server:/var/opt/hawkular
+    - "9990:9990"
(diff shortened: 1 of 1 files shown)


rhqci:

pilhuhn:
There is already a docker-compose template in docker-dist/ before accepting this PR we should figure out what to do with that template, so that there is only one place to maintain.

jotak:
Ah right, I searched for "docker-compose.yml" and the other one is named differently. Is there a specific reason for not naming it docker-compose.yml?

pilhuhn:
I guess I did it that way to illustrate it, but not to mandate how to do it (so people could/should modify to match their local need)

jotak:
Ok. Then it's up to you for what we do about this PR. An advantage I would see in renaming the file, is that it makes it clear that it can be used as is, whereas with the current name my first idea is that it's a template file used for some build step in the maven process and I wouldn't be sure it can be used unless I have a closer look at it. (I admit it's not a big deal, except for lazy people :) )

pilhuhn:
@jshaughn wdyt?
@jotak If Jay agrees, then update your pr to remove the -template one.

jshaughn:
@pilhuhn @jotak, I think we should just have a clear docker-compose.yml file.  No one will be using this stuff other than us, it's not so much for public consumption but rather for what will make our lives easier.

jshaughn:
@jotak, so do you want to update the PR and also fix the license issue?  We can then probably move forward and merge.

jotak:
@jshaughn done

jshaughn:
Thanks @jotak, I created HAWKULAR-1235 for this task.

PR #169, opened 2017-07-07
Supports avail-change Notification

josejulio:
Hawkular-services now listen for the `AvailChange` Notification.
It will react on this notification for Server types, creating an event to inform the avail change (e.g. to track Loss of connectivity).

Still needs to plug something onto BackfillCacheManager to close a hole, as the server won't notify when is down because the agent will also be down.


rhqci:

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/ListenerUtils.java
```diff
@@ -48,39 +55,62 @@ public ListenerUtils() {
...
         try {
             init();
-
             String tenantId = resourcePath.ids().getTenantId();
             eventId = (null == eventId || eventId.isEmpty()) ? UUID.randomUUID().toString() : eventId;
 
             if (checkExists) {
-                if (null != alerts.getEvent(tenantId, eventId, true)) {
-                    return;
+                if (miqTargetId == null) {
```
jshaughn: @josejulio Can you better explain the need for the targetId/familyId stuff?  Is it not possible to generate a unique eventId for each avail change?  The reason we have to deal with duplicate events for events like ResourceAdded is because the agent can rediscover the same resource on restarts.  But I don't see why that would be a problem for an availability change. It seems you could just report each change for the targeted resources.  Let me know what you think because these additions add a decent amount of complexity, let's make sure it is required.  Maybe @jmazzitelli has a comment here.

josejulio: I guess is easier to explain with an example. 

Suppose the following scenario:

1. Agent discovers Deployment `A`, it checks it avail: `UP` and sends notifications, `Deployment A is UP`. Services is notified and adds a new event with ID `Deployment A is UP`. 
2. Avail check, Deployment `A` avail changes `DOWN and sends notification, `Deployment A is DOWN`. Services is notified and adds a new event with ID `Deployment A is DOWN`
3. Avail check, Deployment `A` avails changes to `UP` and sends notification, `Deployment A is UP`. Services is notified and tries to add a new event with ID `Deployment A is UP`. Error id `Deployment A is UP` is already used.

Thus i though on adding a new label to group together same kind events to the same target (e.g.  `avail changes of Deployment A`) and append an UUID to each eventId. (`Deployment A is UP_bb6e4828-e74a-4f70-87ce-98acf0fe3efc`) to avoid duplicates.

jshaughn: Why not just make the event id a UUID or something like that?   The reason the "ResourceAdded" event ids are predictable is so that we can detect a duplicate event.  But if we don't think avail events will be duplicates then I don't think the value of the eventId matters as long as it's unique.

josejulio: Though on that, but what if the agent restarts and previous event was about `Deployment A is UP`, upon restarting it would still send `Deployment A is UP` again and services would create a new event.
Furthermore, I  though on doing the same with the "ResourceAdded" (for the server leaving domain or something) with a "ResourceRemoved". Being able to track if it gets added / removed and avoid to duplication of the event.

If you don't think this needed i could remove that extra logic and keep it simple.

jshaughn: > ... upon restarting it would still send Deployment A is UP again

I don't think we should report the initial avail, only a change.  In that approach I don't think we need to worry about duplicates.  I would encourage a more simple approach for now.  The only reason we handled duplicates for ResourceAdded was because we knew for sure it would be an issue. @jmazzitelli , any thoughts?

jmazzitelli: Not sure how you avoid sending on "initial avail" but only send on change. The agent collects avails without knowing what the state was before - its just "time to check avail - collect avail - store avail". Unless I'm missing something. It's been a while since I looked at the avail check code.

jshaughn: I believe @josejulio now caches the previous value. So I expect the previous value would be null or something like that, on the initial report.

jshaughn: That caching is needed just to know whether there was a change of avail.  The main issue here is whether we need to add this extra complexity to avoid duplicate events, or if we can just continue to drive off of the eventIds.

josejulio: That's right, I already have a cache and is possible to know if is the first avail (`previousAvail == null`)

josejulio:
@jshaughn I added support on the backfill mechanism for the server avail. 
I moved logic to do specific processing of the Notifications on a MIQEventUtil class that can be accessed from the Backfill code without needing to know about Notifications.

I still need to figure a way to send an Avail UP. I suppose services would need to do that, because the agent can't do that as the current availchange mechanism won't detect any aviail change on a local server from down to up.

Please let me know your thoughts on that.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
```diff
@@ -396,11 +402,24 @@ private void doBackfill(CacheKey key, CacheValue value) {
...
             MetricId<AvailabilityType> metricId = new MetricId<>(key.getTenantId(), MetricType.AVAILABILITY,
-                    invMetric.getId());
-            String monitoringType = (String) invMetric.getProperties().get(MONITORING_TYPE_KEY);
+                    invMetric.get().getId());
+            String monitoringType = (String) invMetric.get().getProperties().get(MONITORING_TYPE_KEY);
             List<DataPoint<AvailabilityType>> availList = MONITORING_TYPE_VALUE_REMOTE
                     .equalsIgnoreCase(monitoringType) ? unknown : down;
             Metric<AvailabilityType> backfillAvail = new Metric<>(metricId, availList);
+
+            if (invMetric.getParent() instanceof org.hawkular.inventory.api.model.Resource.Blueprint) {
```
jshaughn: Can you add an inline comment as to why we have this side-effect/conditional logic in the backfill logic.  It won't be obvious to someone in the future.   Also, let me make sure I understand what you are doing here.  The "if" tests to see if we are at the top of the hierarchy, assuming that the topmost avail metric represents the server hosting the embedded agent.  And then it generates an avail change event for that host server. Yes?

Assuming that I'm correct, a couple of comments.  Just above this we determine whether we're setting the avail to UNKNOWN or DOWN based on` monitoringType`.  Although we don't really expect REMOTE monitoring type, we should limit our conditional to only generate the availChange event for a DOWN avail (in that way we confirm this is an embedded agent).  And therefore, we should be reporting a DOWN avail change, not UNKNOWN like you are using below.

Hopefully this makes sense and I'm thinking about this correctly.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
```diff
@@ -396,11 +402,24 @@ private void doBackfill(CacheKey key, CacheValue value) {
...
+                CanonicalPath resourcePath = CanonicalPath.of()
+                        .tenant(key.getTenantId())
+                        .feed(key.getFeedId())
+                        .resource(resource.getId())
+                        .get();
+                miqEventUtils.handleResourceAvailChange(resourcePath.toString(), invMetric.get().getName(),"UNKNOWN");
+
+            }
+
             return backfillAvail;
```
jshaughn: unnecessary newline

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/InventoryHelper.java
```diff
@@ -64,6 +65,51 @@ private InventoryHelper() {
     }
 
     /**
+     * Helper class to carry additional data with the Blueprint.
```
jshaughn: I guess this was the only way to determine which avail metric was at the top (and therefore the server?).  I thought the cononical path carried the resource type info.

jshaughn:
>  still need to figure a way to send an Avail UP...

Perhaps @jmazzitelli can offer his opinion on what would be the easiest approach.  I was thinking that in the same way you **don't** report the initial avail for all other resources (the cached current avail=null) that you **do** report UP avail for the host server avail metric, although I'm not totally sure how you would know which metric represents the host server.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/InventoryHelper.java
josejulio: I couldn't find any CanonicalPath in the inventory metrics, only `relative path`, thus i needed to add a way to carry additional information with the blueprint (in this class the resource that this metric belongs to)

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
josejulio: > Can you add an inline comment as to why we have this side-effect/conditional logic in the backfill logic. It won't be obvious to someone in the future. Also, let me make sure I understand what you are doing here. The "if" tests to see if we are at the top of the hierarchy, assuming that the topmost avail metric represents the server hosting the embedded agent. And then it generates an avail change event for that host server. Yes?

Not exactly.
This will test if the "parent" is a resource. it should always be, metrics are always children of a resource AFAIK. Still testing to prevent Class cast exceptions.

As is, it will generate the avail change event for every resource that has avail type = `Server Availability`.
This would be ideal if we want to report all servers in  a domain configuration. Is that something we want to do or should I also test for being root level resource?

> Assuming that I'm correct, a couple of comments. Just above this we determine whether we're setting the avail to UNKNOWN or DOWN based onmonitoringType. Although we don't really expect REMOTE monitoring type, we should limit our conditional to only generate the availChange event for a DOWN avail (in that way we confirm this is an embedded agent). And therefore, we should be reporting a DOWN avail change, not UNKNOWN like you are using below.

I'll take into account `onmonitoringType` to set UNKNOWN / DOWN. As `javaagent`s can only monitor "REMOTE" servers (even if they are local).


> Hopefully this makes sense and I'm thinking about this correctly.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
josejulio: Removing, thanks

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/cache/BackfillCacheManager.java
```diff
@@ -396,11 +402,31 @@ private void doBackfill(CacheKey key, CacheValue value) {
             // Set UNKNOWN for all remotely monitored avail metrics reported by this feed/tenant
             // Set DOWN for all locally monitored avail metrics, or by default, reported by this feed/tenant
             MetricId<AvailabilityType> metricId = new MetricId<>(key.getTenantId(), MetricType.AVAILABILITY,
-                    invMetric.getId());
-            String monitoringType = (String) invMetric.getProperties().get(MONITORING_TYPE_KEY);
+                    invMetric.get().getId());
+            String monitoringType = (String) invMetric.get().getProperties().get(MONITORING_TYPE_KEY);
             List<DataPoint<AvailabilityType>> availList = MONITORING_TYPE_VALUE_REMOTE
                     .equalsIgnoreCase(monitoringType) ? unknown : down;
```
jshaughn: OK, thanks for the explanations so far.  After thinking about this, I think the root of the problem is that the javaagent has made some of this legacy backfiller logic obsolete.  The older, embedded, agent had local-dmr and remote-dmr options for monitoring the local/host server or  a remote server, respectively.   The javaagent is basically only using remote-dmr.  But it is, at least currently, only used to monitor the one server with which it is deployed.  So, what happens is we end up with "monitoringType == MONITORING_TYPE_VALUE_REMOTE" despite the fact that logically we're dealing with resources from a local server.

So, what we need to do is the following, remove the logic regarding monitoringType, add a comment mentioning that  we are *assuming* that 1) all resources/metrics are from a javagent and 2) that the javaagent is deployed with and reporting on only one server.  Then, have the backfiller only set DOWN avail, never UNKNOWN.  If things change in the future, and break the assumptions, we'll deal with it going forward.

josejulio: > Perhaps @jmazzitelli can offer his opinion on what would be the easiest approach. I was thinking that in the same way you don't report the initial avail for all other resources (the cached current avail=null) that you do report UP avail for the host server avail metric, although I'm not totally sure how you would know which metric represents the host server.

On the agent, the metrics do have a link to the resource they belong to, I could test if the resource is actually a server and report. 
However that will be less generic, maybe I can use other notification  (e.g. `AVAIL_STARTING`) and put it on the servers types.

josejulio:
I tested workflow and is working well for services + java agent.

@jshaughn  Can I have another review round?

This depends on https://github.com/hawkular/hawkular-commons/pull/108 and https://github.com/hawkular/hawkular-agent/pull/347

josejulio:
Updated commons with src-dep, not sure why that `default` is failing.

jshaughn:
I also don't know why the "default" build is failing and I can't visit the details site.  You did create a hawkular jira for this, if I remember.  Can you update the PR title to mention the jira and/or make sure this PR is linked to the Jira?  If you are confident I will merge the PR despite that off "default" build failure.

josejulio:
Just linked.
I'm not sure if the src-dep made it fail. 
It works well on my local box and passes the tests, so we can go ahead and merge.

jshaughn:
OK, merging despite the weird "default" build failure for which we can't even see the details.  If there is an issue we'll see it soon enough and then fix it.

PR #170, opened 2017-07-17
Version of Origin Metrics Cassandra on Openshift Yaml Template

gbaufake:
Reviewing Hawkular Services Openshift Images and I saw that
Hawkular-services:0.38 was released  ~ a month ago  and  origin-metrics-cassandra v1.4.1 was released ~ 6 months ago.

I tested with origin-metrics-cassandra v1.5.1 (which was released ~ 2 months ago) on openshift cluster and it seems to be working.


rhqci:

Jiri-Kremser:
@gbaufake lgtm, but could you also please change it [here](https://github.com/hawkular/hawkular-services/blob/ebda0c9b82076553998b1eb6f1df648d90468ebd/openshift/hawkular-services-ansible-vars.yaml#L21)?

PR #171, opened 2017-08-01
Bumps commons to 0.9.7.Final

josejulio:


rhqci:

PR #178, opened 2017-09-12
Initial commit for MWM++ architecture

lucasponce:
- Basic layout using HawkularServer (Vert.X based server)
- Add Commons and Inventory projects


lucasponce:
@jshaughn @jotak @jmazzitelli

jotak:
@lucasponce I didn't realize that you were adding pieces of commons here. Since I'm modifying commons in the same time it's out of sync.
So the idea is to get everything about mwm poc in a single repo?

lucasponce:
@jotak well, yes, the HawkularServer is and other common utils are hosted in commons and in this initial attempt I used them just to replicate the same stack we added in alerting.
I see the issue, so, perhaps we might put everything under the single repo for poc and we can move code later.
I think that in the current discussions perhaps the inventory model API will be only used from the hServices, so yes, it could make sense.

lucasponce:
@jshaughn could you review this ? @jotak would need it  to consolidate things

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/MIQEventUtils.java
```diff
@@ -1,77 +0,0 @@
...
-import java.util.HashSet;
-import java.util.Set;
-
-import org.hawkular.inventory.paths.CanonicalPath;
-import org.hawkular.listener.bus.ListenerUtils;
-
-/**
- * Group together MIQ events sent to alerting
- */
-public class MIQEventUtils {
```
jshaughn: TODO : What happens in this class will still be important going forward.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/CommandEventListener.java
```diff
@@ -1,90 +0,0 @@
...
- * payload should be a command pojo.  We want to generate only one Hawkular Event per Command Gateway event, so this
- * is Queue based, limiting message consumption to one server.
- * </p>
- * @author Jay Shaughnessy
- */
-@MessageDriven(messageListenerInterface = MessageListener.class, activationConfig = {
-        @ActivationConfigProperty(propertyName = "destinationType", propertyValue = "javax.jms.Queue"),
-        @ActivationConfigProperty(propertyName = "destination", propertyValue = "HawkularCommandEvent") })
-@TransactionAttribute(value = TransactionAttributeType.NOT_SUPPORTED)
-public class CommandEventListener extends BasicMessageListener<BasicMessage> {
```
jshaughn: TODO : What happens in this class will still be important going forward.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/ListenerUtils.java
```diff
@@ -1,103 +0,0 @@
...
-import java.util.UUID;
-
-import javax.naming.InitialContext;
-
-import org.hawkular.alerts.api.model.event.Event;
-import org.hawkular.alerts.api.services.AlertsService;
-import org.hawkular.inventory.paths.CanonicalPath;
-import org.jboss.logging.Logger;
-
-public class ListenerUtils {
```
jshaughn: TODO : What happens in this class will still be important going forward.

PR #179, opened 2017-09-14
Init InventoryService with backend/ispn stub

jotak:
Needs this [1] in order to compile.
For the time being hawkular-services doesn't have dependency management in parent pom, I don't know if it's intentional... 

[1] https://github.com/hawkular/hawkular-services/pull/180

### DIFF
--- a/inventory/pom.xml
+++ b/inventory/pom.xml
@@ -32,14 +32,25 @@
   <name>Hawkular Services: Inventory</name>
 
   <dependencies>
+    <dependency>
+      <groupId>org.hawkular.commons</groupId>
+      <artifactId>hawkular-inventory-model</artifactId>
+      <version>${version.org.hawkular.commons}</version>
+    </dependency>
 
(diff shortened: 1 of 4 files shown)


lucasponce:
I wonder if we can consolidate the inventory API in hawkular-services too.
It seems we won't need a separate component as potential agent code could live here too.
What do you think @jotak ?
Then in the first steps we can avoid to release commons or using srcdeps.

rhqci:

jotak:
@lucasponce +1 I'll move the model

jotak:
Closing, I'll open another one

jshaughn:
> I wonder if we can consolidate the inventory API in hawkular-services too.
It seems we won't need a separate component as potential agent code could live here too.

That has been my assumption all along, but to make it clear, yes. I thing Inventory will be a native part of hServices/MWM.  If the "agent" is also central/native then interaction will be local.

PR #181, opened 2017-09-14
Init InventoryService with backend/ispn stub + REST API

jotak:


rhqci:

lucasponce:
I am +1 to migrate.
We have discussed to migrate this functionality into a .war inside the former hServices architecture.
I can prepare that.
Minor question for me is if to create a mwm-wf branch that we can use to consolidate, then this branch can be used for poc purposes as the other one would need a little bit more of plumbing and glue code.

jotak:
@lucasponce or for going on with two branches (actually I think mwm-wf will become the only active one but anyway it doesn't cost to create branches)

PR #182, opened 2017-09-27
Fix 'dict object' has no attribute 'stdout' when oc_client_before_15 …

rubenvp8510:
…is registered with skipped.


rhqci:

PR #183, opened 2017-09-27
Updating dependencies

theute:
Definitely need some verification.
I blindly updated to latest versions, did a build and started it.


rhqci:

gbaufake:
@theute is that ok, if we can wait Hawkular Services 1.6.3.Final https://github.com/hawkular/hawkular-metrics/pull/894 to be merged on Hawkular Metrics 0.26.x line?

This pull request include a correction for ManageIQ use case when it queries for events.

PR #184, opened 2017-09-29
Preliminar package for MWM++

lucasponce:
- Remove old depdencies from metrics, old inventory jars and cassandra
- Remove unnecessary/outdate jms configs
- Remove temporarily Backfill feature and derivates as this one might be implemented through P8S

There are still things to clean and work, but basically the hServices is working with SRCDEPS to put in a single place the new inventory, alerting, cmdgw and listeners in a single place:

```
GET /hawkular/inventory/status
{"Implementation-Version":"0.9.8.Final-SRC-revision-767ee6ffeb784fa49c1ad4ba0203aa4426705166","Built-From-Git-SHA1":"767ee6ffeb784fa49c1ad4ba0203aa4426705166","status":"UP"}

GET /hawkular/alerts/status
{"distributed":"false","Implementation-Version":"1.8.0.Final-SRC-revision-a32591c387011a64975fc3a3338c0a79df2a5ac4","Built-From-Git-SHA1":"a32591c387011a64975fc3a3338c0a79df2a5ac4","status":"STARTED"}
```

Distribution starts and stops without errors.
ITests pass but there is an error on JMS queue when stopping (I am looking at it).

Openshift and Docker distribution are not yet touched.

Agent is not yet touched.

Please @jotak @jmazzitelli @jshaughn take a look.


lucasponce:
> This PR potentially fails with hawkular-client-ruby.

Of course, it does

jotak:
Nice purge. I won't comment all the feature pack changes as I'm ignorant in that field. The rest looks good to me.

jshaughn:
> Remove temporarily Backfill feature and derivates as this one might be implemented through P8S

+1, it is correct to remove the avail backfill code as there are no avail metrics going forward, Instead there is the built-in UP TS provided by P, per instance, based on whether it can scrape.  I expect that will be our only avail.   There is the possibility that we may want to keep the agent websocket connection listener.  We may find that we'll want to know if the agent goes down.

feature-pack/src/main/xsl/subsystem-templates/hawkular-services-logging.xsl
```diff
@@ -30,15 +30,6 @@
       <logger category="org.hawkular.alerts">
         <level name="${{hawkular.log.alerts:INFO}}" />
       </logger>
-      <logger category="org.hawkular.metrics">
-        <level name="${{hawkular.log.metrics:INFO}}" />
-      </logger>
-      <logger category="org.hawkular.inventory">
-        <level name="${{hawkular.log.inventory:INFO}}" />
-      </logger>
```
jshaughn: Do we not want an inventory logger entry?

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/MIQEventUtils.java
```diff
@@ -39,37 +38,32 @@
...
-    public void handleResourceAdded(String resourceType, String resourcePath) {
+    public void handleResourceAdded(String resourceType, String feedId, String resourceId) {
         final String RESOURCE_ADDED = "RESOURCE_ADDED";
-        CanonicalPath cp;
-        try {
-            cp = CanonicalPath.fromString(resourcePath);
-        } catch (Exception e) {
-            throw new IllegalArgumentException(
-                    "Required Property [resourcePath] is missing or is an invalid CanonicalPath: " + e.getMessage());
+        if (feedId == null || resourceId == null) {
```
jshaughn: Maybe we could add a couple of simple utils like isEmpty(feedid, resourceId) { return isEmpty(feedid) || isEmpty(resourceId) }

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/MIQEventUtils.java
```diff
@@ -39,37 +38,32 @@
...
-            cp = CanonicalPath.fromString(resourcePath);
-        } catch (Exception e) {
-            throw new IllegalArgumentException(
-                    "Required Property [resourcePath] is missing or is an invalid CanonicalPath: " + e.getMessage());
+    public void handleResourceAvailChange(String feedId, String resourceId, String availType, String newAvail) {
+        if (feedId == null || resourceId == null) {
+            throw new IllegalArgumentException("Required Property [feedId] or [resourceId] is missing.");
+
         }
+
```
jshaughn: It's fine to keep this for now but we will probably have to revisit this feature impl.  I'm not sure the agent will be able to provide this information going forward, as it no longer collects avail.  Instead we may need to add something alerting-based.  I'll think about that today.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/HawkularTopicListener.java
```diff
@@ -1,62 +0,0 @@
...
-        // Feed/Agent WebSocket Closed Handling
-        //
-        if (msg instanceof FeedWebSocketClosedEvent) {
-            FeedWebSocketClosedEvent fce = (FeedWebSocketClosedEvent) msg;
-            log.debugf("Feed WebSocket Closed. feedId=%s reason=%s code=%s", fce.getFeedId(), fce.getReason(),
-                    fce.getCode());
-            backfillCacheManager.forceBackfill(fce.getFeedId());
-        }
-    }
-}
```
jshaughn: As mentioned in a PR comment.  We may want to reintroduce this listener based on evolving use cases.  But I don't think it would be used to trigger a backfill, only perhaps to register an agent down event, or something like that.

feature-pack/src/main/xsl/subsystem-templates/hawkular-services-logging.xsl
lucasponce: Yes, but it is provided from commons.
Having this one here is redundant (and makes the config fails).

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/MIQEventUtils.java
lucasponce: +1 ok, will add that.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/MIQEventUtils.java
lucasponce: +1 on this case I simply move the resourcePath into feedId and resourceId.
I can leave it or remove it if you prefer, no real preference from my side.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/HawkularTopicListener.java
lucasponce: Aggred.
I think we start from the point that P8S will provide this info, but if use case changes we can re-introduce it again.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/MIQEventUtils.java
jshaughn: leave it for now, I guess.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/bus/HawkularTopicListener.java
jshaughn: P8S can't provide this info, I think, unless the agent writes its own "UP" metric.  Either way it can be removed now and added back later if we see a need to know when the agent has disconnected.

lucasponce:
> This PR potentially fails with hawkular-client-ruby.

Of course, it does

jshaughn:
@lucasponce Good stuff.

PR #186, opened 2017-10-03
this installs the new javaagent, while leaving some of the old agent around

jmazzitelli:
The old agent isn't installed/running, but the old download and installer zip files are available. The new javaagent is installed and running.


jmazzitelli:
do not merge this yet. it is here to test the new agent on the new inventory. At this time, it appears there are problems with the agent's inventory rest requests.

rhqci:

jotak:
I've built it with current hawkular-agent and, at least for the inventory REST API, it seems to work well. I haven't tested the command gateway.

If someone is interested, I've pushed the corresponding docker image on my account: `jotak/hawkular-services:0.40.0.Final-SNAPSHOT`

jmazzitelli:
no more SNAPSHOT dep. Can someone merge this?

PR #192, opened 2017-10-12
Able to specify custom SSL certificates using secrets.

rubenvp8510:
@Jiri-Kremser  could you please review?

// cc: @theute


rhqci:

theute:
run tests
(This is to test Jenkins pipeline, but it won't do anything interesting on that PR)

theute:
Looking at Travis results, looks like we need a new version of the parent for mycila

theute:
this is ok to test
(Another attempt at kicking Jenkins)

theute:
So we need 2 different ones (+1 for ephemeral ?)

What are the prerequisite (I don't know how to test)
   - Create 2 PVs, 1 for Cassandra, 1 for Hawkular Services
   - Create 1 secret linked to 1 service account ?

@hhovsepy could you have a look ?

rubenvp8510:
For that example, you can just use the playbook.yaml

ansible-playbook playbook.yaml --ask-sudo-pass

I set the hawkular_use_ssl variable to true by default, it create the secrets and the service account, I think those things are the prerequisites for run the template.

openshift/hawkular-services-persistent-ssl.yaml
```diff
@@ -0,0 +1,279 @@
...
+- apiVersion: v1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: h-services-pvc
+  spec:
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 1Gi
```
hhovsepy: historically hardcoed value, please use CASSANDRA_DATA_LIMIT

openshift/hawkular-services-persistent-ssl.yaml
```diff
@@ -0,0 +1,279 @@
...
+- apiVersion: v1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: cassandra-pvc
+  spec:
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 1Gi
```
hhovsepy: HAWKULAR_SERVICES_DATA_LIMIT

rubenvp8510:
@hhovsepy I've modified the scripts to use `HAWKULAR_SERVICES_DATA_LIMIT` and `CASSANDRA_DATA_LIMIT`, please take a look.

theute:
@rubenvp8510 I wanted to try on my proper OpenShift 3.6 installation, without cheating with the Ansible playbook and oc cluster up ;) But I guess I can find my answers in that playbook

theute:
The playbook is not a deliverable, we can make this in a different PR.
What we need before merging, is an update to the parent it seems... (because of mycila stuff)

rubenvp8510:
run tests

rubenvp8510:
Not sure why Jenkins fail.

rhqci:

jpkrohling:
Sorry if I'm stating the obvious here, but perhaps it would ring a bell.

```
15:59:30 FAILED CONFIGURATION: @BeforeSuite arquillianBeforeSuite
15:59:30 org.jboss.arquillian.container.spi.client.container.LifecycleException: The server is already running! Managed containers do not support connecting to running server instances due to the possible harmful effect of connecting to the wrong server. Please stop server before running or change to another type of container.
15:59:30 To disable this check and allow Arquillian to connect to a running server, set allowConnectingToRunningServer to true in the container configuration
...
...
15:59:30 java.lang.Throwable: Method AlertingITest.postGetDelete()[pri:0, instance:org.hawkular.services.rest.test.AlertingITest@3c19aaa5] depends on not successfully finished methods
15:59:30 java.lang.Throwable: Method MetricsAlertingITest.createTriggerSendMetricsReceiveEvents()[pri:0, instance:org.hawkular.services.rest.test.MetricsAlertingITest@38d8f54a] depends on not successfully finished methods
...
```

Basically, it's failing because a test is trying to start a Wildfly instance but there's a Wildfly instance up and running already. If your test expects a server to be up before it starts, then you might need to add the mentioned property to the Arquillian configuration. If you don't expect a Wildfly instance to be up when your test starts, then a previous test might have failed to bring it down cleanly.

rubenvp8510:
@jpkrohling Thank you!, but I'm wondering why start failing with this PR because I haven't move anything related to wildly or the code or the tests? 

Or perhaps this fails for other PRs also?

jpkrohling:
Looks like it's not about your PR, as other PRs seem to have failed as well (build 152).

rubenvp8510:
Is this  ok to merge?

jshaughn:
(11:51:45 AM) rubens: jshaughn, green light to merge :) I will add other improvements but will be in another PR.

PR #197, opened 2017-10-16
use agent with res config

jmazzitelli:


rhqci:

PR #205, opened 2017-10-25
Uniform Hawkular Services names on openshift templates

rubenvp8510:


rhqci:

jshaughn:
@rubenvp8510 Consistent naming is good.  There are other changes in here regarding delay times an failure thresholds.  They are not mentioned in the comment and there is no BZ.  Can you add something to explain why they are updated?

rubenvp8510:
@jshaughn I took those values from downstream templates, I'm not really sure what are good values here, I'll remove those changes from this PR as I'll investigate why those values are different in downstream templates.

openshift/hawkular-services-persistent.yaml
```diff
@@ -53,23 +53,19 @@ parameters:
 - name: ROUTE_HOSTNAME
   description: Under this hostname the Hawkular Services will be accessible, if left blank a value will be defaulted.
   displayName: Hostname
-- name: HAWKULAR_USE_SSL
-  description: Should the self-signed certificate be created
-  displayName: Use SSL
-  value: "true"
```
theute: Why is that section gone ?

openshift/hawkular-services-persistent.yaml
```diff
@@ -85,10 +81,6 @@ objects:
       port: 8080
       protocol: TCP
       targetPort: 8080
-    - name: https-8443-tcp
-      port: 8443
-      protocol: TCP
-      targetPort: 8443
```
theute: Same comment as above

openshift/hawkular-services-persistent.yaml
```diff
@@ -146,8 +138,6 @@ objects:
             value: remote
           - name: CASSANDRA_NODES
             value: hawkular-cassandra
-          - name: HAWKULAR_USE_SSL
-            value: ${HAWKULAR_USE_SSL}
```
theute: And same

theute:
I think I have my answers, I guess this is just some cleanup in the non-SSL template

rubenvp8510:
Well, I think I gonna reintroduce this sections, because of @mwringe  comments on IRC.

jshaughn:
@rubenvp8510 , when you and @mwringe are happy I can merge this.

rubenvp8510:
For me it's ok to merge now.

PR #206, opened 2017-10-26
Hawkular 1259

jshaughn:
This merges the 1259 (new inventory) work into master.  It is set to DNM until we have a green light to do the branching.  Also, we may want to make a release of commons and the agent, and then update this to use the new release deps.


jshaughn:
This should now be ready for merge if it goes green in Travis.  It does have src-deps for agent and commons, both of which will need releases before we can release a "1259/NewInv" version from master.

PR #208, opened 2017-10-28
use hawkular.rest.url since it is a url not a hostname

jmazzitelli:
hawkular.rest.host has been changed to ...url. We also did this in the agent repo as well where hawkualr.rest.host existed, too. This makes things consistent across the board.


rhqci:

PR #212, opened 2017-10-30
Update docker-assembly with removed files

jotak:


rhqci:

docker-dist/src/main/docker/docker-assembly.xml
```diff
@@ -64,11 +64,6 @@
       <fileMode>755</fileMode>
     </file>
     <file>
-      <source>src/main/resources/check-cnode.sh</source>
```
jmazzitelli: remove this check-cnode.sh as part of this PR, it isn't used anymore so should be deleted.

jotak: It's actually already gone, and that's why docker:build failed

jmazzitelli:
while you a fixing the docker image, this should also be cleaned up:

https://github.com/hawkular/hawkular-services/blob/hawkular-1275/docker-dist/src/main/docker/Dockerfile#L33

and related stuff in here (METRICS_TTL env var also should be removed and anything else)

PR #228, opened 2017-11-08
Add possibility to specify SSL public/private key path using parameters.

rubenvp8510:
This is for 0.40.x


rhqci:

josejulio:
LGTM

PR #229, opened 2017-11-08
Support SSL using internal openshift CA

rubenvp8510:
@mwringe Could you please review?

  - Removed SSL Version.
  - Use Service Serving Certificate Secrets
  - Removed parameter SSL enable/disable


rhqci:

mwringe:
Do we know how this will affect current clients? The agent is now going to have to know about the ca certificate used to sign Hawkular Services

rubenvp8510:
@mwringe 

I'm not sure how the customers are affected by this. if they are not using SSL this doesn't have any effect.

But I don't know how they used SSL with this template now,  the current version just generated a self-signed certificate, and store the certificate and private key in the container, the certificate wasn't exposed using some external volume or secret, so as far as I know there wasn't a way to point  the agent to the certificate or a CA to trust on it. ( I don't know if it has some configuration to skip that validation)

But anyway I'll do some tests on agents. Any other observation or issue on this?

// cc @theute @hhovsepy

theute:
We don't have to keep backward compatibility here

openshift/hawkular-services-ephemeral.yaml
```diff
@@ -176,6 +180,9 @@ objects:
         volumes:
         - name: h-services-data
           emptyDir: {}
+        - name: hawkular-secret-volume
+          secret:
+            secretName: hawkular-ssl-secret
```
hhovsepy: Can this value be represented as Parameter with default value instead of hardcoding?

openshift/hawkular-services-ephemeral.yaml
```diff
@@ -138,16 +135,23 @@ objects:
           - name: CASSANDRA_NODES
             value: hawkular-cassandra
           - name: HAWKULAR_USE_SSL
-            value: ${HAWKULAR_USE_SSL}
+            value: "true"
```
hhovsepy: general question, do we need here default SSL support for Ephemeral template?
@rubenvp8510 @theute 
My intense here was to remove SSL for Ephemeral at all.

rubenvp8510: Ok, I'll remove, I used it for some tests.

openshift/hawkular-services-ephemeral.yaml
rubenvp8510: +1

hhovsepy: so based on comment below, all the above SSL and cert related code should be gone

openshift/hawkular-services-ephemeral.yaml
hhovsepy: ok thank you

openshift/hawkular-services-persistent.yaml
```diff
@@ -186,6 +191,9 @@ objects:
         - name: h-services-data
           persistentVolumeClaim:
             claimName: h-services-pvc
+        - name: hawkular-secret-volume
+          secret:
+            secretName: hawkular-ssl-secret
```
hhovsepy: here is still required to make parameter

openshift/hawkular-services-persistent.yaml
```diff
@@ -53,9 +53,6 @@ parameters:
 - name: ROUTE_HOSTNAME
   description: Under this hostname the Hawkular Services will be accessible, if left blank a value will be defaulted.
   displayName: Hostname
-- name: HAWKULAR_USE_SSL
-  description: Should the self-signed certificate be created
-  displayName: Use SSL
   value: "true"
```
hhovsepy: please remove this remaining value of removed property

openshift/hawkular-services-persistent.yaml
```diff
@@ -78,6 +75,7 @@ objects:
     annotations:
       description: Exposes and load balances the application pods
       service.alpha.openshift.io/dependencies: '[{"name":"hawkular-cassandra","namespace":"","kind":"Service"}]'
+      service.alpha.openshift.io/serving-cert-secret-name: 'hawkular-ssl-secret'
```
hhovsepy: here as well, please use property

openshift/hawkular-services-persistent-ssl.yaml
```diff
@@ -1,279 +0,0 @@
...
-  kind: Route
-  metadata:
-    name: ${ROUTE_NAME}
-  spec:
-    host: ${ROUTE_HOSTNAME}
-    to:
-      kind: Service
-      name: hawkular-services
-    port:
-      targetPort: https-8443-tcp
```
hhovsepy: [1] shouldn't it to be moved to next yaml file?

openshift/hawkular-services-persistent-ssl.yaml
```diff
@@ -1,279 +0,0 @@
...
-    selector:
-      name: hawkular-services
-    strategy:
-      type: Rolling
-    template:
-      metadata:
-        labels:
-          name: hawkular-services
-      spec:
-        serviceAccountName: ${SERVICE_ACCOUNT_NAME}
```
hhovsepy: [2] should not it be keps in next yaml file?

hhovsepy:
I did compared old and new templates, and found differences [1] and [2] which are in comments.

openshift/hawkular-services-ephemeral.yaml
mwringe: why do we want to make this secret name configurable?

openshift/hawkular-services-persistent.yaml
mwringe: why should it be a property?

openshift/hawkular-services-ephemeral.yaml
rubenvp8510: Well, I was thinking about this and found that is this name is not relevant.

The only use case that I thought is to avoid collisions if there is another secret with this name on the same namespace? But I think that is unlikely to happen.

openshift/hawkular-services-persistent-ssl.yaml
rubenvp8510: No, we don't need this service account anymore.

rubenvp8510: That is because we are using service certificate secrets[1], so OpenShift will handle all of this.


1)  https://docs.openshift.com/container-platform/3.5/dev_guide/secrets.html#service-serving-certificate-secrets

openshift/hawkular-services-persistent-ssl.yaml
rubenvp8510: We don't create an SSL route by default, the user needs to create it if he wants his own certificate on the "exposed to the world" route.

rubenvp8510:
@hhovsepy I removed SSL support from ephemeral as requested in the review. as  @mwringe I don't see why we need to parametrize the secret, but if we need that I think I can do in another PR. WDYT?

openshift/hawkular-services-persistent.yaml
hhovsepy: because customers creates secret themself before deploying the template.

hhovsepy:
have tried the new template persistence.yaml, it starts in non ssl mode

openshift/hawkular-services-persistent-ssl.yaml
hhovsepy: I mean here you have 8080 port: https://github.com/rubenvp8510/hawkular-services/blob/e3e6e08a234d6d680c2317881cb8f1aedf3313a9/openshift/hawkular-services-persistent.yaml#L121

hhovsepy:
Verified the hawkular-services-persistent.yaml

PR #230, opened 2017-11-08
run h-services and prometheus all in docker

jmazzitelli:
no need to have a local build in target anymore. just give someone run.sh and prometheus.yml and you can run a demo of h-services


rhqci:

PR #231, opened 2017-11-08
use prometheus 2.0

jmazzitelli:


rhqci:

jotak:
LGTM

jmazzitelli:
opps, don't merge yet. I think the yml file is not correct when running out of target

jmazzitelli:
ok, all better now. once green we can merge.

PR #233, opened 2017-11-10
Remove unused Caches and Destinations from config

lucasponce:
Alerting 1.8 manages configuration outside standalone.xml, so those caches are redundant and creates noise also in agent.
Also removing unused destinations, we don't communicate alerting through the bus.


jshaughn:
yes, good cleanup.

PR #240, opened 2017-11-17
fix test due to the new platform resources

jmazzitelli:


rhqci:

PR #241, opened 2017-11-17
HAWKULAR-1336 - configure build to fix issue with dev profile

pgier:
If the "dev" profile is activated, generation of the hawkular services
dist zip is skipped to speed up the build.  The itest and docker-dist
modules are dependent on this zip, so the build will fail if the zip
is not in the local maven repo.  This change causes the itest and
docker-dist modules to be skipped completely if the "dev" profile is active.

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -101,9 +101,6 @@
     <profile>
       <!-- A profile to build a development distro -->
       <id>dev</id>
-      <properties>
-        <dozip>false</dozip>
-      </properties>
 
       <build>
         <plugins>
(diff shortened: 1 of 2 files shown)


jmazzitelli:
NOTE: this PR is against master branch. we would need to merge it into the hawkular-1275 branch as well.

jmazzitelli:
I'm going to merge this. Hawkular-1275 branch doesn't need this necessarily.

For a dev to run the itests but still build a dev server, they now run:

`mvn -Pdev -Pitest -Pdist`

 (note the new -Pdist profile).

PR #245, opened 2017-11-23
Avoiding SSL usage for the agent inside container

rubenvp8510:
There are some cases when I don't have control of the certificates and can't add localhost SANs, an example is when I use OpenShift Service Serving Certificate Secrets.

I think is safe to not use SSL for this case, even if the SSL is enabled. because all traffic is inside the container, which has the private/public keys of the SSL encryption.

This is just for the case when HAWKULAR_SSL=true, so the other way of seeing this is PR is "setting up the SSL but making sure that agent inside the container still use plain HTTP)

### DIFF
--- a/feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template-docker-ssl.xsl
+++ b/feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template-docker-ssl.xsl
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2016-2017 Red Hat, Inc. and/or its affiliates
+    and other contributors as indicated by the @author tags.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


rhqci:

rubenvp8510:
@pgier @theute Could you please review?

feature-pack/src/main/xsl/configuration/standalone/hawkular-services-template-docker-ssl.xsl
```diff
@@ -0,0 +1,53 @@
...
+  <xsl:template match="/*[local-name()='server']/*[local-name()='management']/*[local-name()='security-realms']/*[local-name()='security-realm' and contains(@name, 'ManagementRealm')]">
+    <xsl:element name="security-realm" namespace="{namespace-uri()}">
+      <xsl:attribute name="name">UndertowRealm</xsl:attribute>
+      <xsl:element name="server-identities" namespace="{namespace-uri()}">
+        <xsl:element name="ssl" namespace="{namespace-uri()}">
+          <xsl:element name="keystore" namespace="{namespace-uri()}">
+            <xsl:attribute name="path">hawkular.keystore</xsl:attribute>
+            <xsl:attribute name="relative-to">jboss.server.config.dir</xsl:attribute>
+            <xsl:attribute name="keystore-password">hawkular</xsl:attribute>
+            <xsl:attribute name="key-password">hawkular</xsl:attribute>
```
mwringe: do we want the password to be hardcoded?

rubenvp8510: This is for internal use only, I think it is ok to have it hardcoded, at least for the purposes of this PR.

PR #246, opened 2017-11-23
Add metadata to Persistent Volume example so we can use in web console.

rubenvp8510:


rhqci:

PR #250, opened 2017-11-30
Using gosu for execute the SSL certificate import process.

rubenvp8510:
This is for fix BZ 1515883

https://bugzilla.redhat.com/show_bug.cgi?id=1515883

- First I do all the certificate import process as root,
- then I use `gosu` to switch to `jboss` user and execute the rest of the process (including wildly with hawkular-services deploys)

In this way I avoid all issues with permissions on the volume, This hasn't failed on my machine because I have UID `1000` on my user. (same as `jboss` user on the image), but if the host user UID doesn't match with `jboss` the ownership of the certificates on the volume could be another user (event a non-existing user on the container) and the `jboss` user won't be able to read and import the certificates to the keystore file.

### DIFF
--- a/docker-dist/src/main/docker/Dockerfile
+++ b/docker-dist/src/main/docker/Dockerfile
@@ -22,6 +22,7 @@ MAINTAINER Hawkular Team, hawkular-dev@lists.jboss.org
 ARG agent_enable
 ARG metrics_ttl=21
 ARG the_tag
+ARG GOSU_VERSION=1.10
 
 EXPOSE 8080 8443 8787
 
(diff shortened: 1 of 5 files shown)


rhqci:

docker-dist/src/main/docker/Dockerfile
```diff
@@ -46,7 +47,22 @@ RUN yum install --quiet -y openssl && \
...
+    gosu nobody true; \
+    yum -y remove wget dpkg; \
+    yum clean all
+
 VOLUME ["${HAWKULAR_DATA}"]
 
-USER jboss
-CMD /opt/hawkular/bin/startcmd.sh
+USER root
+CMD /opt/hawkular/bin/entrypoint.sh
```
pgier: Will this still work if the container is started as a non root user?

rubenvp8510:
I'll close this PR, not a good solution for Openshift environments.

PR #251, opened 2017-11-30
Increase the liveness and readiness probes timeout

rubenvp8510:
This PR increases the timeout of liveness probe and readiness probe in order to avoid the following exception:

` org.jboss.resteasy.spi.UnhandledException: RESTEASY003770: Response is committed, can't handle exception connection reset by peer` 

This exception happens because the timeout is too small and the request takes time under some conditions and the probe interrupts the request.


rhqci:

jshaughn:
LGTM

PR #253, opened 2017-12-04
Bug 1424627 Add EAP Host Name to Event Details

jshaughn:


pgier:
I think the bz number is missing a '7' at the end, should be 1424627

jshaughn:
@josejulio Please review.

hawkular-rest/hawkular-listener/src/main/java/org/hawkular/listener/MIQEventUtils.java
```diff
@@ -67,11 +67,19 @@ public void handleResourceAvailChange(String resourcePath, String availType, Str
...
                     "MiddlewareServer", message);
         }
     }
 
+    private String getResource(CanonicalPath cp) {
+        if ( null == cp ) {
+            return "";
+        }
+        String rp = cp.ids().getResourcePath().toString();
+        return null == rp ? "" : ( rp.startsWith("r;") ? rp.substring(2) : rp );
```
josejulio: I might be wrong, but if the server type is e.g. `Domain WildFly Server` it will show the path to the resource instead of the `hostname`, is that OK?

jshaughn: I'm not sure what would be best, I'm willing to make a change.  I thought that maybe if the type were some child type it would be good to show path.  To be honest I can't recall exactly what the old inventory API shows for resourcePath but I believe it can get a 'r;' prefix.   What would you suggest, just showing the final segment of the path (is that the correct term in old inventory?).

jshaughn: I added DNM until we agree on the format.  With the PR as of now we have (for the local server):
Server Added: "Added WildFly Server: Local~~"
Avail Change: "Avail change [Server Availability]: Local~~ [UP]"

So, for a server I think you get the server name, if we were to report on child resources I think you would get the path like Local~~/Child.

@josejulio  The question I guess we need to answer is whether the tildas (~~) need to be removed?  Is there some utility/algorithm used in the gem that I should emulate on the server for formatting the resource?

josejulio: > @josejulio The question I guess we need to answer is whether the tildas (~~) need to be removed? Is there some utility/algorithm used in the gem that I should emulate on the server for formatting the resource?

The provider does this:

https://github.com/ManageIQ/manageiq-providers-hawkular/blob/master/app/models/manageiq/providers/hawkular/inventory/parser/middleware_manager.rb#L398

jshaughn:
OK, @josejulio, I removed the tildas but stopped short of type-specific parsing done in the gem.  I think this is sufficient for a timeline msg and for TP3.  In GA we're using a completely different identifier.  So, I removed the DNM label.

josejulio:
LGTM

PR #258, opened 2017-12-13
HAWKULAR-1347 Fix AgentITest assert result

lucasponce:
This is to set Hawkular Services in a good state.
Once, agent is updated and server is in "running" state again, this test will fail and we will update the result.

### DIFF
--- a/itest/src/test/java/org/hawkular/services/rest/test/AgentITest.java
+++ b/itest/src/test/java/org/hawkular/services/rest/test/AgentITest.java
@@ -64,7 +64,7 @@ public class AgentITest extends AbstractTestBase {
                                                 "[%s] should have a resource with an id of [%s]",
                                                 testResponse.getRequest(), wfServerId));
                                 Assert.assertTrue(dequote(foundResources.get("config").get("Server State").toString())
-                                        .equals("running"),
+                                        .equals("reload-required"),
                                         String.format(
                                                 "[%s] should have a config 'Server State' with value 'running'",
                                                 foundResources.get("config")));


jmazzitelli:
We can do this, or we have the agent run a reload after it updates the statistics enable flags (which probably is what we should do - this "reload-required" tells me the statistic enable flag changes didn't take effect

lucasponce:
In any solution, my point was to have the itests running.
When agent is updated, if this value is again in "running" we can change it.

jmazzitelli:
k - when this goes green, we can merge.

itest/src/test/java/org/hawkular/services/rest/test/AgentITest.java
```diff
@@ -64,7 +64,7 @@ public void agentDiscoverySuccess() throws Throwable {
                                                 "[%s] should have a resource with an id of [%s]",
                                                 testResponse.getRequest(), wfServerId));
                                 Assert.assertTrue(dequote(foundResources.get("config").get("Server State").toString())
-                                        .equals("running"),
+                                        .equals("reload-required"),
```
josejulio: Shouldn't we better fix it to have `running` as the `Server State`? I recall we use this value inside MIQ.

lucasponce: As I put in comment, yes. But that requires a bump in the agent.
But to not leave Hawkular Services in a broken state, let's fix this now and change when agent is ready.

josejulio: Sorry, didn't see that, sorry

lucasponce:
Merging this to not block Travis.
I am not sure if @jmazzitelli worked in update the agent to return into "running" state (performing a "reload").
I am going to look into that as well.
/cc @jshaughn @josejulio just to coordinate our tasks.

